PR #2, opened 2026-02-11
fix(config): user config at ~/.config/kaku/kaku.lua is never loaded on macOS

yikZero:
## Summary

User config files (`~/.config/kaku/kaku.lua`) are never loaded on macOS because the bundled `Resources/kaku.lua` has higher priority in the config search path.

## Problem

In `config/src/config.rs`, the macOS bundled config path is inserted at **position 0** via `paths.insert(0, ...)`, making it the first candidate. Since the loading logic is first-match-wins (returns on the first file found), user configs are never reached:

```
Actual loading order:
1. Kaku.app/Contents/Resources/kaku.lua  ← always found first, returned immediately
2. ~/.kaku.lua                           ← never reached
3. ~/.config/kaku/kaku.lua               ← never reached
```

This contradicts the [README](https://github.com/tw93/Kaku#customization) which states:
> 1. **Self-Contained**: `Kaku.app/Contents/Resources/kaku.lua` handles default settings.
> 2. **User Overrides**: Create `~/.config/kaku/kaku.lua` and return your configuration table.

## Fix

Change `paths.insert(0, ...)` to `paths.push(...)` for the macOS Resources path, so the bundled config serves as a **fallback** (lowest priority) rather than an override:

```
Fixed loading order:
1. ~/.kaku.lua
2. ~/.config/kaku/kaku.lua               ← user config, found first if exists
3. Kaku.app/Contents/Resources/kaku.lua  ← fallback for zero-config experience
```

This preserves the zero-config experience (bundled config is used when no user config exists) while allowing user customization as documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,9 @@ Kaku comes with a carefully curated suite of CLI tools, pre-configured for immed
 
 Kaku is fully configurable via standard Lua scripts and is 100% compatible with WezTerm configuration. It loads configuration files in the following priority order:
 
-1. **Self-Contained**: `Kaku.app/Contents/Resources/kaku.lua` handles default settings.
-2. **User Overrides**: Create `~/.config/kaku/kaku.lua` and return your configuration table.
+1. **Explicit Override**: `KAKU_CONFIG_FILE=/path/to/kaku.lua` (if set).
+2. **User Config**: `~/.config/kaku/kaku.lua`.
+3. **Bundled Fallback**: `Kaku.app/Contents/Resources/kaku.lua`.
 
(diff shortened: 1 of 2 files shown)


tw93:
@yikZero Thanks for the catch — you were right.

We fixed the macOS load-order issue so the bundled `Kaku.app/Contents/Resources/kaku.lua` is now fallback only, and user config is loaded first.

We also simplified the user config entry to a single path:
`~/.config/kaku/kaku.lua`
(and dropped `~/.kaku.lua` to avoid ambiguity).

README has been updated accordingly.

PR #13, opened 2026-02-11
feat: add Windows and Linux build support

czyt:
- Add X11, Wayland, and Windows window backends
- Add filedescriptor Windows implementation
- Add Linux build script and CI workflow for all three platforms
- Add Linux procinfo support and noop toast notification fallback
- Fix termwiz cross-platform compilation (TerminalWaker stub for Windows, gate unix-only terminal functions behind #[cfg(unix)])


tw93:
@czyt Thank you so much for this PR and for the effort you put into Kaku.

We reviewed it carefully. Right now, Kaku is still focused on improving the macOS experience first (stability, performance, and defaults), and our near-term roadmap remains macOS-first.

This PR introduces a large cross-platform surface (build/release pipeline, window system, platform assets), which is broader than what we can take on at this stage.

So for now, we won’t merge this PR (we’ll close it for the moment). Once the macOS core is more mature, we’ll revisit cross-platform support in a more systematic way, and we’d be very happy to have your help then as well.

Thanks again for your contribution and understanding.

PR #17, opened 2026-02-11
fix(shell): respect user's default shell instead of hardcoding zsh

yamamel:
## Summary
- **kaku.lua**: Use `$SHELL` environment variable for `config.default_prog`, falling back to `/bin/zsh` when unset
- **first_run.sh**: After first-run setup, launch the shell by priority: user config (`config.default_prog`) > `$SHELL` > `/bin/zsh`, with executable validation

## Test plan
- [ ] Verify fish/bash users get their own shell instead of zsh
- [ ] Verify zsh users are unaffected
- [ ] Verify fallback to `/bin/zsh` when `$SHELL` is unset
- [ ] Verify first_run.sh respects `config.default_prog` in user's `kaku.lua`

### DIFF
--- a/assets/macos/Kaku.app/Contents/Resources/kaku.lua
+++ b/assets/macos/Kaku.app/Contents/Resources/kaku.lua
@@ -291,7 +291,12 @@ config.colors = {
 }
 
 -- ===== Shell =====
-config.default_prog = { '/bin/zsh', '-l' }
+local user_shell = os.getenv('SHELL')
+if user_shell and #user_shell > 0 then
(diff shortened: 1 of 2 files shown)


tw93:
Thanks for the PR. I can confirm the underlying issue is real (there are currently hardcoded zsh paths).

After reviewing the changes, I found an edge case in the current implementation: the first-run/upgrade shell resolution can still fall back to /bin/zsh under the default config.

To avoid regressions, I implemented a complete fix directly in this repo and linked it to this PR via commit #17 reference. Really appreciate the report and the proposed direction.

PR #20, opened 2026-02-12
refactor: apply idiomatic Rust cleanups across GUI and CLI

dongowu:
Replace non-idiomatic patterns with Clippy-recommended alternatives: unwrap_or_default, Copy dereference over clone, implicit returns, bool negation, is_empty, usize::MAX, and bare unwrap to expect.


tw93:
LGTM

PR #22, opened 2026-02-12
fix(gui): correct split pane rendering and text wrapping

yamamel:
## Summary
- **Clip underlines to visible content area**: underlines were rendering beyond the pane's visible pixel width. Added bounds check to break early, matching existing background clipping logic.
- **Adjust pane terminal size for split padding**: the GUI adds visual padding around split lines, but the terminal was unaware of the reduced area, causing text to wrap too late. On resize, each pane's column count is now reduced by the padding columns consumed by adjacent splits.

## Test plan
- [ ] Open a split pane layout (vertical splits) and verify text wraps at the correct visible boundary
- [ ] Confirm underlines in split panes don't extend beyond the pane edge
- [ ] Test with different `window_padding` configurations
- [ ] Verify single-pane (no split) behavior is unchanged

### DIFF
--- a/kaku-gui/src/termwindow/resize.rs
+++ b/kaku-gui/src/termwindow/resize.rs
@@ -297,6 +297,45 @@ impl super::TermWindow {
         if let Some(window) = mux.get_window(self.mux_window_id) {
             for tab in window.iter() {
                 tab.resize(size);
+
+                // Adjust each pane's terminal size to account for split padding.
+                // The GUI adds visual padding around split lines, but the terminal
+                // emulator must wrap text earlier to match the reduced visible area.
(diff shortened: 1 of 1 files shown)


tw93:
@yamamel Thanks for the contribution! We merged the underline clipping fix in screen_line.rs,that's a clean and correct fix.

For the resize.rs change, we've reverted it. The core issue is that window_padding.left represents the padding around the whole window edge, not the split bar width. The split bar itself is only 2px (split_thickness = 2.0 in split.rs) and doesn't need column compensation. Using window padding to calculate column reduction could cause incorrect text wrapping even in normal (non-split) layouts with non-zero padding.

If you're seeing actual text wrapping issues in split panes, happy to investigate the root cause together!

PR #34, opened 2026-02-13
fix: restore double-click to zoom with synchronous drag

GavinRiver:
## Summary

Restores double-click-to-zoom functionality for the title bar and padding area, using a synchronous drag approach that avoids the text selection issue from #24.

### Changes

- **`mouseDownCanMoveWindow` returns `NO`** — ensures `mouseDown:` reaches the app layer so we can detect double-clicks and handle drag ourselves
- **Synchronous drag via `PENDING_DRAG_MOVE`** — `request_drag_move()` sets a thread-local flag; `mouse_down` checks it after `mouse_common()` returns and calls `performWindowDragWithEvent:` synchronously (inspired by [Ghostty's approach](https://github.com/ghostty-org/ghostty/blob/main/src/apprt/embedded.zig))
- **Double-click detection in title/padding area** — uses existing `last_mouse_click.streak == 2`, returns early to skip drag initiation

### Why this won't re-introduce #24

The previous approach (`mouseDownCanMoveWindow = YES`) caused macOS to intercept **all** mouseDown events, preventing the app from distinguishing clicks from drags. This fix returns `NO` so every mouseDown reaches the app layer, where upstream's existing protections work correctly:

1. `is_window_dragging` flag
2. `current_mouse_capture` (set to `UI` during title-area drag)
3. `mouse_event_terminal` checks capture before selecting text
4. macOS `performWindowDragWithEvent:` enters a modal drag loop that doesn't generate move events to the view

### Test plan

- [x] Single-click drag in title area → window moves
- [x] Double-click title area → window maximizes
- [x] Double-click when maximized → window restores
- [x] Double-click tab bar empty area → window maximizes/restores
- [x] Click in terminal area → no interference with text selection
- [x] Drag in terminal area → text selection works normally
- [x] Window buttons (close/minimize/maximize) → work correctly

### DIFF
--- a/kaku-gui/src/termwindow/mouseevent.rs
+++ b/kaku-gui/src/termwindow/mouseevent.rs
@@ -155,9 +155,15 @@ impl super::TermWindow {
                 // Perform click counting
                 let button = mouse_press_to_tmb(press);
 
+                // 标题/padding 区点击用哨兵行值，防止和终端首行(row=0)串联成双击
+                let click_row = if event.coords.y < terminal_origin_y {
+                    i64::MIN
+                } else {
+                    y
+                };
                 let click_position = ClickPosition {
                     column: x,
-                    row: y,
+                    row: click_row,
                     x_pixel_offset,
                     y_pixel_offset,
                 };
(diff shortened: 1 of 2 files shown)


tw93:
LGTM

PR #66, opened 2026-02-16
feat: add TogglePaneSplitDirection to toggle split H↔V

GavinRiver:
## Summary

Add `TogglePaneSplitDirection` action that toggles the split direction of the active pane's parent split node between Horizontal (left|right) and Vertical (top|bottom).

- **Shortcut**: `Cmd+Shift+S`
- Similar to iTerm2's "Swap Split Direction", but neither iTerm2, Ghostty, nor upstream WezTerm have this as a single-key toggle

## How it works

1. Find the active pane in the split tree
2. Move up to its parent split node
3. Flip `H ↔ V` and recalculate child dimensions (50/50 split)
4. Cascade sizes down to nested children via `cascade_size_from_cursor`

## Edge cases handled

| Case | Behavior |
|------|----------|
| Single pane (no split) | No-op, returns immediately |
| Pane is zoomed | No-op, returns immediately |
| Insufficient space for new direction | No-op, silently skipped (avoids spurious resize) |
| Nested splits (3+ panes) | Only flips the active pane's immediate parent split |

## Files changed (5 files, +127 lines)

- `config/src/keyassignment.rs` — enum variant
- `mux/src/tab.rs` — core toggle logic + minimum-size guard
- `kaku-gui/src/termwindow/mod.rs` — action dispatch
- `kaku-gui/src/commands.rs` — command definition
- `assets/.../kaku.lua` — keybinding (`Cmd+Shift+S`)

## Test plan

Tested manually + automated AppleScript on macOS:

- [x] Single pane → `Cmd+Shift+S` → no change, no crash
- [x] H-split → toggle → becomes V-split → toggle → back to H-split
- [x] V-split → toggle → becomes H-split
- [x] Nested 3-pane → toggle → only active pane's parent flips
- [x] Zoomed pane → toggle → no change; unzoom → layout preserved

### DIFF
--- a/assets/macos/Kaku.app/Contents/Resources/kaku.lua
+++ b/assets/macos/Kaku.app/Contents/Resources/kaku.lua
@@ -676,6 +676,13 @@ config.keys = {
     action = wezterm.action.TogglePaneZoomState,
   },
 
+  -- Cmd+Shift+S: Toggle split direction (horizontal <-> vertical)
+  {
+    key = 'S',
+    mods = 'CMD|SHIFT',
+    action = wezterm.action.TogglePaneSplitDirection,
+  },
+
   -- Cmd+Ctrl+Arrows: Resize panes
   {
     key = 'LeftArrow',
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -633,6 +633,7 @@ pub enum KeyAssignment {
 
     CopyMode(CopyModeAssignment),
     RotatePanes(RotationDirection),
+    TogglePaneSplitDirection,
     SplitPane(SplitPane),
     PaneSelect(PaneSelectArguments),
     CharSelect(CharSelectArguments),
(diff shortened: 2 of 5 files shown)


GavinRiver:
Hi @tw93, the CI failure is not related to code formatting — it's a checkout issue with the new CI workflow.

The `fmt` job uses `ref: ${{ github.head_ref }}`, which tries to find the branch `feat/toggle-split-direction` in `tw93/Kaku`, but since this is a fork PR, the branch only exists in `GavinRiver/Kaku`.

Previous PRs (#34, #36) passed CI because they used the old workflow (with "Format Check" instead of "Format").

I've verified locally that formatting is clean:
```
cargo +nightly fmt --check -p kaku -p kaku-gui -p mux -p config
# No issues found
```

A simple fix for the CI would be to remove the `ref:` line from the checkout step (GitHub Actions defaults to the merge commit for PRs), or use `ref: ${{ github.event.pull_request.head.sha }}`.

tw93 (commented):
LGTM

PR #75, opened 2026-02-17
feat: add per-pane encoding support (UTF-8, GBK, GB18030, Big5, EUC-KR, Shift-JIS)

liu-kan:
This PR adds core per-pane encoding support:

- PaneEncoding enum with 6 encoding variants and dynamic menu ordering
- Config option default_encoding for new panes
- Pane encoding conversion engine (input encoder, output decoder)
- Per-pane encoding storage with AtomicU8 thread-safe access
- Encoding propagation through spawn, split, SSH, and tmux
- Shell menu direct encoding switching
- Launcher submenu with back navigation (Esc returns to parent)
- Works for both local and remote (SSH) terminals
- Split panes inherit encoding from source pane
- Restructure pane encoding as submenu with MRU ordering
- Comprehensive tests for all encodings with edge cases

### DIFF
--- a/mux/src/pane_encoding.rs
+++ b/mux/src/pane_encoding.rs
@@ -412,4 +412,309 @@ mod tests {
         let text = decode_bytes_to_string(PaneEncoding::Gbk, &gbk_bytes);
         assert_eq!(text, "你好".to_string());
     }
+
+    #[test]
+    fn preserves_escape_sequences_all_encodings() {
+        let encodings = [
+            PaneEncoding::Gbk,
+            PaneEncoding::Gb18030,
+            PaneEncoding::Big5,
+            PaneEncoding::EucKr,
+            PaneEncoding::ShiftJis,
+        ];
+        let csi = b"\x1b[31m";
+        let osc = b"\x1b]0;title\x07";
+        let dcs = b"\x1bPpayload\x1b\\";
+        let csi_9b = [0x9b, b'3', b'1', b'm'];
+
+        for enc in encodings {
+            let mut decoder = PaneOutputDecoder::default();
+            assert_eq!(decoder.decode(enc, csi), csi.to_vec(), "{enc:?} CSI");
+            let mut decoder = PaneOutputDecoder::default();
+            assert_eq!(decoder.decode(enc, osc), osc.to_vec(), "{enc:?} OSC");
+            let mut decoder = PaneOutputDecoder::default();
+            assert_eq!(decoder.decode(enc, dcs), dcs.to_vec(), "{enc:?} DCS");
+            let mut decoder = PaneOutputDecoder::default();
+            assert_eq!(
+                decoder.decode(enc, &csi_9b),
+                csi_9b.to_vec(),
+                "{enc:?} CSI 0x9b"
+            );
+        }
+    }
+
+    #[test]
+    fn mixed_text_and_escape_all_encodings() {
+        // GBK: "你" = 0xc4e3, "好" = 0xbac3
+        {
+            let mut decoder = PaneOutputDecoder::default();
+            let mut data = vec![0xc4, 0xe3];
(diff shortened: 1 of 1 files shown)


tw93:
Hi @liu-kan,

Thank you so much for this well-crafted PR! The implementation quality is excellent - comprehensive tests, clean architecture, and thorough integration across SSH/tmux/splits.

I can see you put significant effort into this.

Before merging, I'd like to better understand the real-world use cases. Could you share:

1. **What scenarios require per-pane encoding?** For example, connecting to legacy GBK/Big5/Shift-JIS systems?
2. **Why can't server-side encoding conversion solve this?** (e.g., `export LANG=zh_CN.UTF-8` on the remote system)
3. **Have you received user requests for this feature?** Or is this based on your own workflow needs?

Kaku is positioned as a modern terminal for AI coding workflows, so I want to make sure we're adding features that genuinely benefit the user base rather than increasing complexity for edge cases.

Looking forward to your thoughts!

liu-kan:
Hi @tw93 

Thank you for the thoughtful review and for raising these important questions. I appreciate the opportunity to clarify the context behind this feature.

**Primary Use Case: Legacy System Maintenance**

The main scenario involves connecting to and maintaining legacy database systems that simply cannot be upgraded. These are often critical business systems running on internal VPN networks. Thinking the scenario that Oracle instances deployed 15 years ago, or the risk of migration outweighs the benefits. In these environments, database schemas, table names, and even column content are encoded in legacy character sets.

**Why Server-Side Conversion Isn't Viable**

People think `export LANG=zh_CN.UTF-8` would solve this, but unfortunately, it's not an option for these specific legacy systems. Many were designed before UTF-8 became popular, and some were intentionally built for GB18030 compliance (China's official standard) rather than Unicode. The database engines themselves store metadata in these encodings, and the client connections expect the same. Forcing UTF-8 at the SSH session level results in messy code when querying system tables or viewing historical data with Chinese characters. We don't control these systems, they're black boxes we must interface with, or the risk of migration no one wants to touch. There are infrastructures that we cannot modernize.

**Demand & Current Workarounds**

This need comes from both my own workflow (maintaining enterprise legacy integrations) and observations across the enterprise side developer and operate & maintain  community. The pain is real, but the free tools are stuck in the past:

- Many developers still resort to **PuTTY** (last updated 2019, limited modern features, and not MacOS version)
- **Bitvise** is closed-source and Windows-only
- **WindTerm** is promising but partially open-source and heavily focused on Serial connections, and lacks the AI-centric workflow Kaku provides

These engineers aren't choosing ancient tools because they prefer them—they're the only ones that handle encoding switches gracefully per session. As sometime you need to mainten both UTF-8 and regional/localized encodings system, someone even run a Windows VM on macos or remotely to use this feat. There's a quiet but significant community of developers maintaining "digital archaeology" on these systems who've been waiting for a modern terminal that acknowledges this reality.

**Alignment with Kaku's Vision**

When Kaku launched as a lightweight, fast terminal, I saw the opportunity: Mac lacked an open-source, free option supporting legacy GBK encodings with true cross-platform potential. Honored to contribute this feature.

While Kaku targets modern workflows, real-world engineering involves maintaining both cutting-edge infrastructure and legacy systems daily. It will be good if engineers don't need to switch between windterm for old database servers and a modern terminal for everything else—per-pane encoding allows Kaku to serve as the **unified daily driver** for all terminal work.

The implementation respects the UTF-8-first design: the feature adds minimal UI surface (a discreet dropdown or setting), leaves default behavior untouched, and introduces zero friction for standard users. It simply removes the forced context-switching between tools without compromising Kaku's streamlined experience for the majority.

I used WezTerm for a while but eventually moved on. Kaku clearly represents a significant improvement over existing terminals. Thank you for bringing such a polished and intuitive tool to the community.

PR #76, opened 2026-02-17
feat: support branch switch

Binaryify:
This PR add branch switch support:
1. Support opening the branch list modal by clicking the current branch name, and switching branches via mouse or keyboard.
2. Support filtering the branch list by keyword.
The UI screenshot:
image

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -1769,6 +1769,10 @@ fn default_initial_cols() -> u16 {
 
 pub fn default_hyperlink_rules() -> Vec<hyperlink::Rule> {
     vec![
+        // Treat prompt git-branch segments as clickable.
+        // Examples: ` main`, ` main`, ` main`.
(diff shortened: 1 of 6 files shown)


tw93:
@Binaryify Thanks for the PR. After reviewing, we’re leaning toward not merging this for now:

1) The current behavior only injects git switch/checkout into the terminal, which doesn’t really improve speed or reliability.  
2) There’s a non-trivial risk of accidental command injection while the pane is busy or when the worktree is dirty, and failure feedback is weak.  
3) We want to keep Kaku focused as a terminal rather than growing toward a Git client.

If we revisit this later, we’d likely consider a lower-risk approach (e.g., copy branch name, or insert but don’t execute, plus stronger state checks). 
 
Thanks again for the contribution.

Binaryify:
@tw93 Thanks for your reply. I just found this feature in Warp and thought it was very useful.

PR #81, opened 2026-02-17
feat(shell): auto-detect 1Password SSH agent and add IdentitiesOnly

WongLoki:
## What

Auto-detect 1Password SSH agent in Kaku's shell integration and
add `-o IdentitiesOnly=yes` to prevent "Too many authentication failures".

## Why

1Password's SSH agent offers **all** stored keys to the server.
When key count exceeds `MaxAuthTries` (default 6), the connection
is rejected before reaching the correct key.

## How

Upgraded the ssh() wrapper in setup_zsh.sh to detect 1Password
agent via `SSH_AUTH_SOCK` path and automatically inject
`IdentitiesOnly=yes`. Users can opt out with
`KAKU_SSH_SKIP_1PASSWORD_FIX=1`.

### DIFF
--- a/assets/shell-integration/setup_zsh.sh
+++ b/assets/shell-integration/setup_zsh.sh
@@ -290,10 +290,42 @@ if [[ -f "\$KAKU_ZSH_DIR/plugins/zsh-syntax-highlighting/zsh-syntax-highlighting
 fi
 
 # Auto-set TERM to xterm-256color for SSH connections when running under kaku,
-# since remote hosts typically lack the kaku terminfo entry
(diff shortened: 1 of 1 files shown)


tw93:
@WongLoki Thanks for the PR! Great idea, merged.

Made a small follow-up cleanup after merging: removed the redundant -o SetEnv=TERM=xterm-256color from extra_opts (the kaku branch already handles TERM via the env-var prefix TERM=xterm-256color command ssh, so the SetEnv option was never actually used), and collapsed the three execution branches into two.

Core 1Password detection logic is kept as-is works well.

PR #82, opened 2026-02-17
Reduce default font size: Retina 17pt→14pt, low-res 15pt→13pt

WongLoki:
## Background

Received multiple feedback reports that the default font size is too large on initial launch, reducing usable terminal space.

## Changes

Reduced the default font sizes while keeping the DPI-based dynamic detection logic:

| Display Type | Before | After |
|-------------|--------|-------|
| Retina (DPI ≥ 150) | 17pt | 14pt |
| Low-res external display (DPI < 150) | 15pt | 13pt |

Users can still adjust font size at any time via `Cmd+=` / `Cmd+-`.

## Note

⚠️ **Non-essential cosmetic change** — please review whether you'd like to merge this.

### DIFF
--- a/assets/macos/Kaku.app/Contents/Resources/kaku.lua
+++ b/assets/macos/Kaku.app/Contents/Resources/kaku.lua
@@ -241,7 +241,7 @@ config.font_rules = {
 
 config.bold_brightens_ansi_colors = false
 -- Auto-adjust font size based on screen DPI.
--- Retina (>=150 DPI): 17px, low-resolution external displays (<150 DPI): 15px.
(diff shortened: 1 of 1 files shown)


tw93:
@WongLoki Thanks a lot for the PR and for improving the default UX.

For now, I’ll keep the current defaults (17pt Retina / 15pt low-DPI), since this DPI-based setup already covers lower-resolution screens and keeps out-of-box readability strong. 

The main reason is that I actually feel that this is the most comfortable for the programmer's eyes.

Users can still adjust quickly with Cmd+= / Cmd+-.

If needed later, we can introduce a “Compact” preset instead of changing the global default.

PR #95, opened 2026-02-20
fix: respect user config when XDG config env vars are empty

### LINKED ISSUE #93: [BUG] User config ~/.config/kaku/kaku.lua not loaded on macOS 26
## Description

Changes to `~/.config/kaku/kaku.lua` have no effect. The user config file is not being loaded by Kaku.

## Steps to Reproduce

1. Install Kaku 0.3.0 via `brew install tw93/tap/kakuku`
2. Edit `~/.config/kaku/kaku.lua` to the simplest possible override (as recommended in #31):
   ```lua
   local wezterm = require 'wezterm'
   local config = {}
   config.font_size = 30.0
   return config
   ```
3. Open a new Kaku window (`Cmd + N`)
4. Font size remains at the default — the override is ignored

## Observed Behavior

- Any change to `~/.config/kaku/kaku.lua` (font size, keybindings, color scheme) has no effect.
- Inside a Kaku terminal session, `KAKU_CONFIG_FILE` is set to the **bundled** config path: `/Applications/Kaku.app/Contents/Resources/kaku.lua`.
- The debug overlay (`Ctrl+Shift+L`) does not open, so there is no way to see config errors.

## Expected Behavior

Kaku should load user overrides from `~/.config/kaku/kaku.lua` and apply them on top of the bundled defaults.

## Environment

- **Kaku**: 0.3.0
- **macOS**: 26.2 (Build 25C56)
- **Install method**: Homebrew (`brew install tw93/tap/kakuku`)

## Additional Context

Related to #31. The format recommended there (`local config = {}; config.font_size = 14.0; return config`) also does not work.

A "rename tab" keybinding (`Cmd+Shift+R` via `PromptInputLine`) was also attempted but had no effect, consistent with the config file not being read at all.

mickyyy68:
## Summary
- normalize empty `XDG_CONFIG_HOME` to fallback on `$HOME/.config`
- ignore empty segments in `XDG_CONFIG_DIRS` (for example `::`)
- add focused unit tests for empty, missing, and valid XDG config path inputs
- add a concise inline comment for non-obvious normalization behavior

## Why
Fixes #93 where user config could be ignored and Kaku would fall back to bundled defaults when XDG config env vars were empty/malformed.

## Validation
- `make fmt`
- `make check`
- `make test`
- smoke test with `XDG_CONFIG_HOME=""` confirming loaded config path is `~/.config/kaku/kaku.lua`

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -577,23 +577,96 @@ return config
 "#
 }
 
-fn xdg_config_home() -> PathBuf {
-    match std::env::var_os("XDG_CONFIG_HOME").map(|s| PathBuf::from(s).join("kaku")) {
-        Some(p) => p,
-        None => HOME_DIR.join(".config").join("kaku"),
-    }
+fn xdg_config_home_from(home_dir: &Path, xdg_config_home: Option<OsString>) -> PathBuf {
+    // Normalize empty env values to "unset" to preserve HOME/.config fallback behavior.
+    xdg_config_home
+        .filter(|value| !value.is_empty())
+        .map(PathBuf::from)
+        .unwrap_or_else(|| home_dir.join(".config"))
+        .join("kaku")
 }
 
-fn config_dirs() -> Vec<PathBuf> {
+fn config_dirs_from(
+    home_dir: &Path,
+    xdg_config_home: Option<OsString>,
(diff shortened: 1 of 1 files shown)


tw93:
lgtm

PR #108, opened 2026-02-23
feat(ai): add factory droid config support

nguyenphutrong:
## Summary
- add Factory Droid to Kaku AI tool list and README support matrix
- add a clearer grouped model selector (Factory Provided Models / Custom Models) with model label-to-ID mapping on save
- persist Factory Droid model/reasoning/autonomy to `sessionDefaultSettings`

## Validation
- cargo check -p kaku

image

### DIFF
--- a/kaku/src/ai_config/tui.rs
+++ b/kaku/src/ai_config/tui.rs
@@ -1042,8 +1042,6 @@ fn extract_copilot_fields(val: &serde_json::Value) -> Vec<FieldEntry> {
     fields
 }
 
-
-
 fn extract_factory_droid_fields(val: &serde_json::Value) -> Vec<FieldEntry> {
     let session_defaults = val
         .get("sessionDefaultSettings")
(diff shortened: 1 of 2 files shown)


tw93:
@nguyenphutrong Thanks for the PR and the detailed work.

I took another look and I think `Factory Droid` support is worth adding. It looks like enough people are using it.

That said, I would prefer a simpler first version before merging. The current PR adds a lot of tool specific UI logic, and I want to keep the TUI flow easier to maintain.

I would suggest narrowing the scope like this:

1. Keep `Factory Droid` support in the tool list and config path.
2. Keep reading and writing the basic fields for `Model`, `Reasoning Effort`, and `Autonomy Level`.
3. Make `Model` a free text field.
4. Keep dropdown options only for `Reasoning Effort` and `Autonomy Level`.
5. Remove the hardcoded model catalog and related labels.
6. Remove the grouped selector headers and custom selector flow.
7. Keep the save logic simple and avoid deleting top level keys for now.

I can revisit a richer model picker later if there is strong demand.

Thanks again for the contribution. This is useful and I appreciate the effort.

nguyenphutrong:
@tw93 Thanks for the feedback! I've updated the PR based on your comments.

nguyenphutrong:
Hey @tw93, I noticed something in the recent PR #108 changes that I'd like to clarify.
The is_select_option_selectable function was originally added to support Factory Droid's model selector with grouped sections (e.g., "Factory Provided Models", "Custom Models"). After simplifying to a free-text field, the function now just returns true for all options — meaning all options are selectable.
However, I think this function could be useful in the future if we want to:
- Add "disabled" or "unavailable" states for certain dropdown options
- Show tool-specific opt-in/opt-out behaviors
- Handle conditional options based on other settings
So I've kept it in place rather than removing it entirely.
Question: Would you like me to clean this up (remove the function and simplify move_select_up/move_select_down back to basic index changes), or keep it as-is for potential future use?
Let me know — happy to refactor either way!

tw93:
@nguyenphutrong Thanks for the clarification and for thinking ahead.

I agree this could be useful later, but for this PR I would prefer to remove it and keep the selector logic simple.

Right now `is_select_option_selectable` always returns `true`, so it does not add behavior and only adds extra complexity. Please remove it and simplify `move_select_up` and `move_select_down` back to the basic index changes.

If I need disabled or conditional options later, I can add it back with a concrete use case.

Thanks for checking first.

Issue #112, opened 2026-02-24
[BUG] TERM=kaku forwarded over SSH causes garbled terminal display with zsh plugins


guohai:
## Summary

When using a local Kaku shell tab to SSH into a remote machine, typed commands appear garbled in the display (e.g. ls -lh renders as lss s -lh, exit as exitt). The commands execute correctly —
  only the visual display is wrong. The issue does not occur with iTerm2 or in local Kaku shells.

## Steps to reproduce

1. Open a local Kaku tab (macOS, app bundle installed)
2. Ensure zsh-autosuggestions or zsh-syntax-highlighting is active on the remote machine
3. Run ssh <remote-host> from the local Kaku shell
4. Type any command (e.g. ls -lh, exit)

## Expected vs actual

- **Expected**:
- Commands display correctly as typed.
- **Actual**:
- Characters appear doubled or misplaced in the terminal display. Example frame captured mid-session:

  # guohai@Genie in ~ [0:25:09]
    lss s -lh        ← typed "ls -lh"
  ...
    exitt            ← typed "exit"

## Environment

Please include:
- **Kaku version**: (e.g. `0.4.0`)
- Kaku with bundled app (TERM=kaku default) 0.4.0
- **macOS version**: (e.g. `macOS 15.3`)
- **Machine**: (e.g. MacBook Pro M3)
- **Shell**: (e.g. `zsh`, `bash`, `fish`)
- Remote: Linux with zsh + zsh-autosuggestions / zsh-syntax-highlighting

## Optional details

Add anything helpful if available:
- screenshots or recordings
- relevant config snippet from `~/.config/kaku/kaku.lua`
- debug overlay header and related log lines

You can submit first with minimal info; maintainers may ask follow-up questions.

### Root cause
Kaku sets TERM=kaku in local shells and exports TERMINFO_DIRS pointing to the bundled terminfo. When the user runs the system ssh binary, TERM=kaku is forwarded to the remote machine (standard
  SSH behavior) but TERMINFO_DIRS is not (not in AcceptEnv on most servers). The remote machine has no kaku terminfo entry, so zsh plugins (zsh-syntax-highlighting, zsh-autosuggestions) fall back
  to degraded rendering paths that produce incorrect cursor positioning and character output.

  Kaku's own built-in SSH domain already works around this correctly (mux/src/ssh.rs:271):
  if env.get("TERM").map(|t| t.as_str()) == Some("kaku") {
[...]

PR #113, opened 2026-02-24
fix(shell): override TERM=xterm-256color for ssh sessions

guohai:
When SSHing from a Kaku local shell, TERM=kaku is forwarded to the remote machine but TERMINFO_DIRS is not, so the remote has no kaku terminfo. This causes zsh plugins (autosuggestions, syntax-highlighting) to fall back to broken rendering paths, producing garbled display (e.g. "ls -lh" appearing as "lss s -lh").

Add an ssh wrapper to kaku.sh that applies the same TERM override already used by Kaku's built-in SSH domain (mux/src/ssh.rs:271).

https://github.com/tw93/Kaku/issues/112

🤖 Built with SMT <smt@agora.build>

### DIFF
--- a/assets/shell-integration/kaku.sh
+++ b/assets/shell-integration/kaku.sh
@@ -36,6 +36,15 @@ case "$TERM" in
   ;;
 esac
 
+# Remote machines typically don't have the "kaku" terminfo entry, so when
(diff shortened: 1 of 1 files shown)


tw93:
LGTM!

PR #128, opened 2026-03-01
feat(assistant): support custom request headers

ddnio:
## Summary
- add custom_headers support for Kaku Assistant assistant.toml
- inject custom headers into outbound chat-completions request headers
- keep reserved headers protected (Authorization, Content-Type)
- expose and edit custom headers in kaku ai UI and persist them
- update config template/docs and add tests

## Example
custom_headers = ["X-Customer-ID: demo", "X-Trace-ID: abc123"]

## Notes
- upstream repo is read-only for this account, so this PR is from fork branch ddnio:codex/kaku-assistant-custom-headers


tw93:
@ddnio   Thanks for the contribution! I've merged this PR and made a small follow-up adjustment: the custom_headers option is now a file-only configuration rather than being exposed in the kaku ai TUI.

This keeps the TUI interface simple for most users, while power users who need custom headers for enterprise proxies or API gateways can still configure them directly in ~/.config/kaku/assistant.toml.

I also improved the config template comments to better explain the format and reserved headers.

PR #129, opened 2026-03-01
fix(selection): preserve text selection during active terminal output

GavinRiver:
## Summary

- Remove `check_for_dirty_lines_and_invalidate_selection` which auto-cleared text selections when dirty lines overlapped with selected rows
- Programs with frequent output (status bar animations, streaming logs, spinners) produce dirty lines every frame, making it impossible to select and copy text while output is active
- Selection lifetime is now managed by user actions (click elsewhere, new selection, ESC), consistent with Alacritty/iTerm2/Kitty

## Problem

When running programs that continuously update the terminal (e.g., CLI tools with spinner animations or status bars), any text selection would be instantly cleared on mouse release. The per-frame dirty line check would detect the animation updates as "changed lines" and wipe the selection before the user could copy.

## Changes

| File | Change |
|------|--------|
| `kaku-gui/src/termwindow/mod.rs` | Removed `check_for_dirty_lines_and_invalidate_selection` function (37 lines) |
| `kaku-gui/src/termwindow/render/pane.rs` | Replaced call with explanatory comment |

This also eliminates the per-frame `get_changed_since()` heap allocation overhead from the render loop.

## Test plan

- [x] Run a program with continuous output (e.g., `top`, CLI tools with spinners)
- [x] Select text while output is active → selection persists after mouse release
- [x] Cmd+C copies correctly
- [x] CopyOverlay (Ctrl+Shift+X) and QuickSelect still work (independent seqno tracking)
- [x] Builds with no warnings on macOS

### DIFF
--- a/kaku-gui/src/termwindow/mod.rs
+++ b/kaku-gui/src/termwindow/mod.rs
@@ -2047,43 +2047,6 @@ impl TermWindow {
         }
     }
 
-    fn check_for_dirty_lines_and_invalidate_selection(&mut self, pane: &Arc<dyn Pane>) {
-        let dims = pane.get_dimensions();
-        let viewport = self
-            .get_viewport(pane.pane_id())
-            .unwrap_or(dims.physical_top);
-        let visible_range = viewport..viewport + dims.viewport_rows as StableRowIndex;
-        let seqno = self.selection(pane.pane_id()).seqno;
-        let dirty = pane.get_changed_since(visible_range, seqno);
-
-        if dirty.is_empty() {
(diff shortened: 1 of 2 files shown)


tw93:
LGTM

@GavinRiver Thanks again for the fix and clear write-up.  
I merged it and then did a follow-up cleanup to remove stale selection seqno state that is no longer used after this behavior change. I also removed outdated comments so the selection lifecycle is clearer in code.  

This keeps the implementation consistent and easier to maintain

Issue #106, opened 2026-02-22
[BUG] Tilde is converter to <fffffffff> on some keyboards. (TR in my case)


ragokan:
## Summary

When I press OPT+Tilde key, the text appears as <fffffffff>, when I press enter, I get this error;

<img width="377" height="127" alt="Image" src="https://github.com/user-attachments/assets/ed2a6d57-5713-448c-94ac-9a66fecc125d" />

## Steps to reproduce

Press OPT+Tilde key in keyboard.

## Expected vs actual

- **Expected**: Tilde appearing
- **Actual**: Getting <fffffffff> instead

## Environment

- **Kaku version**: `0.4.0`
- **macOS version**: Tahoe 26.3
- **Machine**: Mac Mini M4 Pro
- **Shell**: `zsh`

## Optional details

My key press flow. I guess it is also in US+Spanish keyboard layout.

<img width="1024" height="464" alt="Image" src="https://github.com/user-attachments/assets/0e67cbb1-1a05-4a7d-a691-c3507efdbe09" />

tw93:
@ragokan Thanks for the report and repro details. This is fixed on our side and will be included in the next release after v0.4.0 (the dead-key placeholder from some macOS layouts like TR was being treated as normal input instead of compose state). Please try the next version when it’s out, and let us know if it still happens.

ragokan:
Hello @tw93 sadly it is still same.

tw93:
@ragokan Thanks for following up. We've extended the dead-key detection to handle keyboard layouts that emit multiple Unicode noncharacters instead of just one. This should fix the <fffffffff> issue on Turkish and similar layouts.

If you'd like to test before the next release, you can build from source:
```
git clone https://github.com/tw93/Kaku.git
cd Kaku
make dev
```
Let us know if it works for you.

ragokan:
Hello, sadly it is still same :( How can I help you to debug it? 

<img width="493" height="231" alt="Image" src="https://github.com/user-attachments/assets/cc7ddf78-99b9-4111-889a-49ca4523091d" />

ragokan:
@tw93 
https://github.com/tw93/Kaku/pull/139/changes

This fixes the issue, maybe you can merge or you can see why it was happening.

Similar issues from Wezterm;

      - Alt/Shift ~ mapping regression: https://github.com/wezterm/wezterm/issues/1706
[...]

PR #139, opened 2026-03-03
macOS: fix tilde typing on Turkish keyboard layouts

ragokan:
## Summary
- fix macOS key translation for layouts that produce `~` via Option/dead-key typing
- when dead-key translation is required (`chars` is empty), force dead-key resolution so typed symbols are emitted correctly
- preserve ASCII punctuation produced by Option chords (eg `~`) when composed-key sending is disabled for that Alt side

## Validation
- `cargo check -p window`
- manual verification: typing `~` and `cd ~` works (copy/paste already worked before)


ragokan:
Issue; https://github.com/tw93/Kaku/issues/106

tw93:
Thanks @ragokan! Your PR is merged. Great catch on the Option chord handling for ASCII punctuation. The fix correctly preserves symbols like ~ that Turkish and similar layouts produce via Option combinations.

ragokan:
Great to hear, thanks!

PR #147, opened 2026-03-05
fix: resolve non-Latin IME blocking Cmd+alnum shortcuts on macOS

zpdldhkdl:
## Summary
- Fix Cmd+alnum shortcuts (e.g. Cmd+W, Cmd+K) not working when a non-Latin IME is active on macOS
- Intercept Cmd+alnum at `performKeyEquivalent` using hardware virtual key codes, bypassing unreliable macOS NSMenu keyEquivalent matching under IME
- Extend support to Cmd+Shift+alnum combinations (e.g. Cmd+Shift+D, Cmd+Shift+A)
- Prevent modifier loss in `key_common` when `unmod` is empty during IME input

## Root Cause
macOS NSMenu `keyEquivalent` matching uses `NSEvent.characters()` internally, which returns incorrect (non-ASCII) characters for some keys when a non-Latin IME is active. This causes certain Cmd+letter shortcuts to fall through to `keyDown`, where the modifier is then stripped due to `chars` vs `unmod` mismatch — resulting in the bare letter being typed into the terminal.

## Changes
- `is_alnum_virtual_key()`: identify A-Z/0-9 by hardware virtual key code
- `is_command_alnum_shortcut()`: match Cmd(+Shift)+alnum, exclude Alt/Ctrl
- `performKeyEquivalent`: intercept Cmd+alnum before NSMenu delegation
- `key_common`: set `prefer_vkey=true` for Cmd+alnum; preserve modifiers when `unmod` is empty

## Test Plan
- [x] `cargo test -p window` — 4 tests passed
- [x] `make fmt-check` — passed
- [x] `make check` — passed
- [x] Manual: non-Latin IME — Cmd+W, Cmd+K, Cmd+Shift+D, Cmd+Shift+A
- [x] Manual: English input — all shortcuts still work (no regression)
- [x] Manual: Cmd+` window cycling preserved

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -2638,6 +2638,27 @@ fn is_symbol_virtual_key(vkey: u16) -> bool {
     .contains(&vkey)
 }
 
+fn is_alnum_virtual_key(vkey: u16) -> bool {
+    [
+        kVK_ANSI_A, kVK_ANSI_B, kVK_ANSI_C, kVK_ANSI_D, kVK_ANSI_E, kVK_ANSI_F, kVK_ANSI_G,
+        kVK_ANSI_H, kVK_ANSI_I, kVK_ANSI_J, kVK_ANSI_K, kVK_ANSI_L, kVK_ANSI_M, kVK_ANSI_N,
+        kVK_ANSI_O, kVK_ANSI_P, kVK_ANSI_Q, kVK_ANSI_R, kVK_ANSI_S, kVK_ANSI_T, kVK_ANSI_U,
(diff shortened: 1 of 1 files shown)


tw93:
LGTM！
@zpdldhkdl Merged and followed up with a regression-hardening commit.

What was added:
- Extracted helper predicates for `performKeyEquivalent` interception and empty-`unmod` modifier handling.
- Added matrix tests for Cmd+alnum interception behavior (including Cmd+Shift variants and non-target modifier combos).
- Added tests to ensure Cmd+alnum does not lose modifiers when `unmod` is empty.

PR #152, opened 2026-03-06
ci: pin bundled shell vendors and validate app bundle

frankekn:
## Summary
- pin bundled zsh plugin vendors to exact commits instead of downloading upstream HEAD
- make the vendor sync idempotent and leave a `.kaku-vendor-ref` marker for verification
- add CI coverage for the real app bundle packaging path used by releases

## Verification
- `bash -n scripts/download_vendor.sh`
- `./scripts/download_vendor.sh && ./scripts/download_vendor.sh`
- `zsh -lc "make app"`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,3 +146,32 @@ jobs:
 
           lipo -info target/universal/release/kaku
           lipo -info target/universal/release/kaku-gui
+
+  app-bundle:
+    name: App Bundle Validation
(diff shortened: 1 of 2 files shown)


tw93:
@frankekn Thanks for calling this out. I fixed this in 242deda so the app-bundle check now validates the pinned vendor refs from inside `dist/Kaku.app/Contents/Resources/vendor` and verifies the expected commit SHAs, which better matches the actual packaged artifact.

PR #159, opened 2026-03-08
fix(gui): improve edge resize cursor and window stability

iwen-conf:
## Summary

- Fix cursor showing I-beam instead of arrow when hovering over window edge resize zones
- Fix `AppDelegate` use-after-free by using process-lifetime singleton pattern instead of `StrongPtr`
- Add macOS live-resize support in WebGPU: read real backing size from `NSView` and switch to `Mailbox` present mode during resize to reduce lag
- Add `edge_drag_in_progress` state to prevent terminal from capturing mouse events during edge resize

## Test Plan

- [x] Build and run `Kaku.app` on macOS
- [x] Hover mouse over window edges — cursor shows arrow (not I-beam)
- [x] Drag window edges to resize — smooth resize without terminal text selection
- [x] Normal text selection in terminal content area still works


tw93:
@iwen-conf This PR currently has real merge conflicts with the latest `main` branch.

I reproduced the conflict locally.Please rebase onto the latest `main` (or merge `main` into this branch) and resolve them on your side.

A few notes for the resolution:

- `.gitignore`: keep both sides. The current `main` additions like `/.sisyphus/` and `/log/` should stay, and your `.ace-tool/` entry should also remain.
- `window/src/os/macos/app.rs`: please keep your AppDelegate lifetime fix. We should not revert back to the old `StrongPtr` approach.
- `kaku-gui/src/termwindow/mod.rs` and `kaku-gui/src/termwindow/mouseevent.rs`: this is not a simple pick-one-side conflict. `main` already added `is_window_dragging` based protection, while this PR adds `edge_drag_in_progress` for resize-zone handling. The final resolution should preserve both behaviors instead of overwriting one with the other.

After resolving, please run `make check` and verify both behaviors manually:

- dragging the window does not trigger terminal selection/scrolling
- resizing from the window edge keeps the correct cursor and does not send unwanted mouse events into the terminal

Once updated, I can take another pass.

tw93:
@iwen-conf Thank you for your contribution.

I resolved the merge conflicts on your branch and pushed a maintainer update to `fix/resize-zone`.

The resolution keeps both behaviors that were needed on top of the latest `main`:
- the AppDelegate lifetime fix on macOS
- the existing `is_window_dragging` protection from `main`
- the `edge_drag_in_progress` handling for edge resize from this PR

I also ran `make check` after the merge. If you don't have additional changes planned here, I'll proceed with the final review/merge from my side.

PR #164, opened 2026-03-09
fix: harden zsh starship RPROMPT fallback

frankekn:
## Summary
- avoid zsh pattern parsing when stripping a literal `$(...)` wrapper from seeded RPROMPT commands
- preserve the existing starship fast path and only harden the fallback path
- extend the smoke test to cover the seeded-command path that previously raised `bad pattern` / `closing brace expected`

## Repro
```sh
zsh -fc 'source ~/.config/kaku/zsh/kaku.zsh >/dev/null 2>&1; RPROMPT="\\$(echo fake-right-prompt)"; _kaku_starship_rprompt_cmd="$RPROMPT"; _kaku_fix_starship_rprompt'
```
Before this patch that path could raise `_kaku_fix_starship_rprompt:13: bad pattern: $(` or `closing brace expected`.

## Testing
```sh
bash assets/shell-integration/tests/starship_rprompt_smoke.sh
```
Passed locally.

### DIFF
--- a/assets/shell-integration/setup_zsh.sh
+++ b/assets/shell-integration/setup_zsh.sh
@@ -389,8 +389,12 @@ if command -v starship &> /dev/null; then
                 if [[ "\${_kaku_starship_rprompt_cmd}" == *starship*'prompt --right'* ]]; then
                     evaled="\$(_kaku_render_starship_rprompt)"
                 else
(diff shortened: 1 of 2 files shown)


tw93:
@frankekn LGTM

PR #171, opened 2026-03-09
feat(i18n): add bilingual UI support (English/Chinese)

iwen-conf:
## Summary

Add bilingual (English / Chinese) UI support to Kaku, controlled by a new `language` config option.

### What's included

**i18n infrastructure** (`config/src/i18n.rs`)
- Lightweight translation layer using a `lazy_static` HashMap
- `t()` / `t_display()` helpers return `Cow<'static, str>` for zero-cost English path
- Supports `"en"` (default) and `"zh"`

**Config integration**
- New `language: Option<String>` field in `Config` struct
- Language is applied on both startup and live config reload
- Persist explicit `color_scheme` selection when changed in Settings TUI

**Translated surfaces**
- Command palette: 60+ command labels wrapped with `t()`
- Launcher overlay: domain/workspace/encoding entries
- Settings TUI: all field labels, section chrome, and footer hints
- Added `unicode-width` dep for correct CJK column alignment in the TUI

### Usage

```lua
-- ~/.config/kaku/kaku.lua
config.language = "zh"   -- Chinese UI
config.language = "en"   -- English UI (default)
```

Or change it in **Settings TUI → Language**.

## Changed Files

| File | Change |
|------|--------|
| `config/src/i18n.rs` | **New** — translation map & helpers |
| `config/src/config.rs` | Add `language` field |
| `config/src/lib.rs` | Register module, init on config load |
| `kaku-gui/src/commands.rs` | Wrap command labels with `t()` |
| `kaku-gui/src/overlay/launcher.rs` | Wrap launcher strings with `t()` |
| `kaku/src/config_tui/mod.rs` | Add Language setting row |
| `kaku/src/config_tui/ui.rs` | Translate Settings TUI chrome |
| `Cargo.toml` / `kaku/Cargo.toml` | Add `unicode-width` dependency |

## Test Plan

- [x] `cargo check` passes
- [x] Settings TUI → change Language to 中文 → all labels switch
- [x] Command Palette shows translated entries in Chinese mode
- [x] `config.language = "zh"` in kaku.lua → Chinese UI on restart
- [x] English remains default when `language` is unset

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2521,6 +2521,7 @@ dependencies = [
  "termwiz-funcs",
  "toml 0.8.23",
  "umask",
+ "unicode-width 0.2.0",
  "url",
  "wezterm-client",
  "wezterm-gui-subcommands",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -241,6 +241,7 @@ uds_windows = "1.1"
 umask = { path = "crates/umask" }
 unicode-normalization = "0.1.24"
 unicode-segmentation = "1.12"
+unicode-width = "0.2"
 url = "2"
 url-funcs = { path = "lua-api-crates/url-funcs" }
 utf8parse = "0.2"
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -194,6 +194,10 @@ pub struct Config {
     /// by the colors setting.
     pub color_scheme: Option<String>,
 
+    /// UI language: "en" for English (default), "zh" for Chinese.
+    #[dynamic(default)]
+    pub language: Option<String>,
+
     /// Named color schemes
     #[dynamic(default)]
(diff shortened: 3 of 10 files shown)


tw93:
@iwen-conf Thanks for the contribution and for putting time into this.

At the moment, we’re not planning to add Chinese UI support to Kaku. I want to keep the product and maintenance surface as small as possible, and introducing UI localization adds ongoing complexity across commands, menus, settings, and future copy changes.

The main issue is not just the initial implementation cost, but the long-term maintenance cost. It is also difficult to guarantee full and consistent coverage across all UI surfaces, and partial localization tends to create an uneven user experience.

For now, I’d prefer to keep the UI in English only and avoid taking on the additional i18n layer. So I’m not planning to move this PR forward in its current direction.

Thanks again for the effort and understanding.

PR #180, opened 2026-03-10
Config padding in cell

inteist:
this is more flexible

added stylua.toml but it didn't really help - had to edit in nvim as VSCode kept messing formatting up.

### DIFF
--- a/assets/macos/Kaku.app/Contents/Resources/kaku.lua
+++ b/assets/macos/Kaku.app/Contents/Resources/kaku.lua
@@ -189,9 +189,9 @@ local low_resolution_screen = is_low_resolution_screen()
 
 local function get_default_padding()
   if low_resolution_screen then
-    return { left = '26px', right = '26px', top = '26px', bottom = '0px' }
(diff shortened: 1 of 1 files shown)


tw93:
@inteist Thanks for the PR.

I’m going to close this for now.

The `window_padding` change is not a neutral refactor. Switching the default from fixed `px` values to `cell` changes the layout semantics and the visual result, because `cell` padding is evaluated from font cell metrics rather than preserving the current fixed spacing. That also weakens the current low-resolution vs high-resolution default spacing split, which is intentional in the bundled config.

On the `stylua.toml` side, we don’t currently use Stylua in the project toolchain or CI, so adding the config alone does not provide a concrete benefit right now.

If we want to revisit either of these, I’d prefer to do it as a separate, intentional change with:
1. a clear goal for the default visual behavior,
2. before/after comparison on different screen classes,
3. toolchain/CI integration for Lua formatting if we decide to adopt Stylua.

Thanks again for the contribution.

PR #181, opened 2026-03-11
Highlight panes with config

inteist:
# Pane Highlight

## Rendering Changes
- Added support for three active pane indicator modes:
  - Bell: circular marker behavior (original)
  - Gutter: left vertical indicator that scales with active pane line usage.
  - Pill: centered, fixed-height left marker.
- Improved geometry calculation to align indicator placement with split-pane layout and content insets.
- Updated leftmost pane handling to avoid off-screen placement and maintain consistent overlap behavior.
- Added helper routines for indicator size and cap radius handling.

## Configuration Changes
- Replaced boolean bell on/off with an enum for active pane indicator mode [off/bell/gutter/pill].
  - active_pane_indicator
  - active_pane_indicator_size
- Added validation for non-zero indicator size.
- Default mode is gutter, with explicit mode enum values for bell/gutter/pill.

## Config TUI Changes
- Moved pane indicator controls under the Panes section.
- Added horizontal left/right adjustment behavior for numeric fields:
  - font_size step: 1
  - line_height step: 0.05
  - active_pane_indicator_size step: 1
- Added visual arrow "hints" around horizontally adjustable values.
- Updated header to include application version: Kaku vX.Y.Z · Settings.
- Improved key/value alignment using a dynamic key column width plus a fixed gap

## Test Coverage
- Added and updated unit tests for:
  - indicator geometry and clamping behavior
  - gutter cap/radius behavior
  - leftmost pane bounds consistency
  - pill centering and size behavior
  - indicator size minimum behavior
  - TUI mode normalization/serialization
  - TUI numeric stepping behavior and limits
  - TUI layout helper behavior for key-column sizing

# Gutter
Screenshot 2026-03-10 at 11 12 39 PM
Screenshot 2026-03-10 at 11 12 33 PM

# Pill
Screenshot 2026-03-10 at 11 13 51 PM
Screenshot 2026-03-10 at 11 14 16 PM

# Config TUI

Screenshot 2026-03-10 at 11 12 57 PM
Screenshot 2026-03-10 at 11 13 20 PM

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -883,9 +883,7 @@ pub struct Config {
     #[dynamic(default)]
     pub audible_bell: AudibleBell,
 
-    /// Chooses the active pane indicator style.
-    /// Bell shows a dot at the top-right corner; Gutter shows a left gutter bar
-    /// tied to active line usage; Pill shows a centered short left pill.
+    /// Active pane indicator style.
     #[dynamic(default)]
     pub active_pane_indicator: ActivePaneIndicator,
 
@@ -2502,6 +2500,7 @@ pub enum NotificationHandling {
 
 #[derive(Debug, ToDynamic, Clone, Copy, PartialEq, Eq, Default)]
 pub enum ActivePaneIndicator {
+    Off,
     #[default]
     Gutter,
     Bell,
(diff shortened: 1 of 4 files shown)


inteist:
P.S. I am very new to Rust, so my sincere apologies if there are some "dumb" things.

tw93:
@inteist Thanks for putting this together. I reviewed the PR in detail, and I don’t think we should merge it in the current form.

The main issue is that it couples two different behaviors:

1. active pane highlighting
2. tab unread-bell indication

Today, tab bell dots are controlled independently by `bell_tab_indicator`. In this PR, that behavior becomes tied to `active_pane_indicator == Bell`, which means switching to `Gutter`, `Pill`, or `Off` also disables tab bell indicators. That is a behavioral regression and changes existing config semantics.

I also think the scope is too large for the value it adds. This PR introduces a new indicator system, a size setting, TUI changes, README updates, and unrelated hook changes, all for what is primarily a visual preference feature. That increases maintenance cost more than I’m comfortable with for mainline.

If you want to continue with this idea, I’d suggest a much smaller follow-up:
- keep `bell_tab_indicator` independent
- make active pane indicator changes purely additive
- avoid changing existing defaults unless there is a strong product reason
- remove unrelated changes from the PR

So the direction is understandable, but I don’t think this version is necessary to merge.

Issue #184, opened 2026-03-11
[FEATURE] keyboard input broadcast to other tab


zxh326:
## Problem

Hello, thank you for creating such an excellent terminal tool. I have been using it to replace my Iterm2 for some time now. The overall experience is quite good.

However, Iterm2 has a feature that allows me to use multiple tabs simultaneously to enter the same commands. This feature is easy to implement in Iterm2 (`command+option+i` ). But I couldn’t find any way to implement this feature in Kaku. Perhaps this is just a feature that isn’t available in Kaku.

## Proposed behavior

What should Kaku do?

## Optional references

You can submit first with minimal info; maintainers may ask follow-up questions.

inteist:
Hey @zxh326, just curious - what is the use case for this? and do you specifically mean tabs or panes?

PR #188, opened 2026-03-11
feat: add pane/tab input broadcast modes

zxh326:
- Toggle All Panes: `Cmd + Shift + I`
- Toggle Current Tab Panes: `Cmd + Option + I`

fix: https://github.com/tw93/Kaku/issues/184

### DIFF
--- a/kaku-gui/src/termwindow/mod.rs
+++ b/kaku-gui/src/termwindow/mod.rs
@@ -4834,8 +4834,21 @@ impl TermWindow {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


tw93 (approved):
@zxh326 LGTM.
I pushed a small follow-up fix for the overlay edge case before approving.

What I changed
- switched the broadcast gate from pane_id equality to pane instance identity
- excluded overlay panes from broadcast targets, so copy/search/quick-select input no longer leaks into real terminal panes
- aligned the visual broadcast state with the same active mux pane check
 
The PR head now includes that fix in commit 2155d3c.

PR #189, opened 2026-03-11
Tui quit without saving

inteist:
minor improvement to “status” line in TUI

### DIFF
--- a/kaku/src/config_tui/mod.rs
+++ b/kaku/src/config_tui/mod.rs
@@ -70,11 +70,15 @@ fn run_app(
 
         match app.mode {
             Mode::Normal => match key.code {
-                // Q / ESC: exit (auto-save if dirty, signal if any save occurred)
-                KeyCode::Char('q') | KeyCode::Char('Q') | KeyCode::Esc => {
+                // ESC: save and exit (auto-save if dirty, signal if any save occurred)
+                KeyCode::Esc => {
                     if let Err(e) = app.save_if_dirty() {
                         return Err(e);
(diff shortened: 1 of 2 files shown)


tw93:
@inteist Thanks for the contribution. There are useful ideas in this PR, especially around the config TUI and active pane indicator.

I’m going to pass on merging it as-is.

The main issue is that it changes existing behavior in ways that are hard to take safely:
- `q` / `Q` now quits without saving, which increases the risk of accidental config loss
- `active_pane_indicator` is now tied to the existing bell tab indicator behavior, which changes defaults and breaks backward compatibility for current configs
- the README and implementation are not fully aligned
- the scope is much larger than the title suggests, which makes review harder

If you’d like to keep working on this, I’d recommend splitting it into smaller PRs with clearer boundaries. A focused PR for “quit without saving” would be much easier to review. The indicator work should also be handled separately, with compatibility preserved for existing configs.

Thanks again for taking the time to put this together.

inteist:
@tw93 of course, no problem. I have realized that this is probably way too much but figured you may like it and roll with it or take some ideas from the PR to implement - this is why I submitted it. I had no illusion that it will most likely be rejected.

Conflating bell and indicators occurred to me but I thought to make the indicator behave similarly like bell (pulsing or similar). 

I have not used Kaku enough to have had experience with the notifications/bell functionality so, initially I completely mistook it for just pane indicator. It was while working on the actual code that I realized my mistake but it was a little too late.

Thanks for your time and your work!

PR #190, opened 2026-03-11
Transparency adjustment

inteist:
- Fixed rendering of the window frame (overlapping drawing created weird effects on the frame)
- TUI matches the transparency
- minor TUI functionality and UX improvements

### DIFF
--- a/kaku-gui/src/termwindow/render/pane.rs
+++ b/kaku-gui/src/termwindow/render/pane.rs
@@ -290,10 +290,7 @@ impl crate::TermWindow {
                     layers,
                     0,
                     background_rect,
-                    palette
-                        .background
-                        .to_linear()
-                        .mul_alpha(window_opacity),
+                    palette.background.to_linear().mul_alpha(window_opacity),
                 )
                 .context("filled_rectangle")?;
             quad.set_hsv(if pos.is_active {


tw93:
@inteist Thanks for the work here. I can see the motivation, especially around the transparency/frame rendering issue and the TUI polish.

I don’t think we should merge this PR in its current form, mainly because it bundles several separate changes together and a couple of them change existing behavior in ways that look risky.

The main concern is that `bell_tab_indicator` currently controls unread bell dots on inactive tabs, but this PR effectively replaces that behavior with `active_pane_indicator`. That means the existing bell indicator behavior is no longer independently configurable, and the default behavior would change as well. I think that is a regression.

I’m also not comfortable merging the rendering changes together with the config model changes, TUI interaction changes, Makefile updates, and version bump script. Even if parts of this are useful, the scope is too broad to review and validate confidently as one PR.

My preference would be:
1. Split out the transparency/frame rendering fix into a minimal PR.
2. Keep `bell_tab_indicator` as a separate setting with its current meaning.
3. Send active pane indicator / config TUI / tooling changes as separate follow-up PRs if needed.

If you want, I’m happy to review a smaller PR focused only on the transparency rendering fix.

PR #214, opened 2026-03-18
fix(gui): require Command key to open links (#212)

### LINKED ISSUE #212: [BUG] Single clicks open files leads to unexpected open
## Summary

What happened? A short description is enough.

When having a file link, a single click opens the file in an editor, which can lead into missclicks opening them while selecting text for instance.
It would be a lot more convenient to keep this feature but having it under ⌘+click rather than just a click

## Steps to reproduce

1. With a file path in the terminal
2. A single click opens the file

## Expected vs actual

- **Expected**: ⌘+click opens a file
- **Actual**: click opens a file

## Environment

Please include:
- **Kaku version**: (e.g. `0.4.0`) 0.7.1
- **macOS version**: (e.g. `macOS 15.3`) 26.3.1
- **Machine**: (e.g. MacBook Pro M3) MacBook Pro M1
- **Shell**: (e.g. `zsh`, `bash`, `fish`) zsh

s010s:
Fixes #212 

## Description
This PR modifies the default mouse click behavior in the terminal to prevent unintended file or link openings. Previously, a single left click without any modifier keys would both complete the text selection and open the link under the cursor. This design caused significant friction because users often single-click to focus the terminal, dismiss selections, or start selecting text. If the cursor happened to be over a file path or URL, it would unexpectedly open it in the editor.

## Changes Made
- Downgraded `Modifiers::NONE` (and other generic combinations like `Modifiers::SHIFT`) from `CompleteSelectionOrOpenLinkAtMouseCursor` to `CompleteSelection`. Now, plain clicks strictly handle selection and focusing, restoring the expected UX of traditional terminal emulators.
- Added a dedicated mapping for `Modifiers::SUPER` (the Command ⌘ key on macOS). Users can now intentionally open links or files by holding the Command key while left-clicking, which aligns with industry standards (e.g., iTerm2, VS Code, macOS Terminal).

### DIFF
--- a/kaku-gui/src/inputmap.rs
+++ b/kaku-gui/src/inputmap.rs
@@ -181,7 +181,7 @@ impl InputMap {
                         streak: 1,
                         button: MouseButton::Left
                     },
-                    CompleteSelectionOrOpenLinkAtMouseCursor(
+                    CompleteSelection(
                         ClipboardCopyDestination::ClipboardAndPrimarySelection
                     )
                 ],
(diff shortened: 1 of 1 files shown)


tw93:
@s010s Thanks for the PR, and for taking the time to work on #212.

I’ve already merged a more complete fix into `main`, so I won’t merge this PR. The final fix also covers the default/bundled config path to make sure the behavior is consistent out of the box.

Appreciate the contribution and the clear description.

PR #222, opened 2026-03-20
Add a setting for basename-only tab titles

rookie-ricardo:
## Summary
- add a `tab_title_show_basename_only` toggle for auto-generated tab titles
- expose the option in the settings panel as `Basename-only Tab Titles`
- document the config override in the default user config template

## Motivation
Kaku currently auto-generates tab titles from the last two path segments, for example `CodeSpace/erduo-site`.

That works well for short paths, but it becomes noisy when the parent folder is long, such as iCloud or deeply nested workspace directories. In that case the tab bar spends a lot of horizontal space on path context that is often less important than the current folder name itself.

This option gives users a simple built-in way to reduce tab title length and keep the tab bar easier to scan when they work across many long-named directories.

## Screenshot
- I will add before/after screenshots in the PR thread.
image

image

## Testing
- `make check`
- `cargo test -p kaku config_tui`

## Notes
- `make fmt` could not run in this environment because `cargo +nightly fmt` requires `rustup`, which is not installed here
- `make test` could not run in this environment because `cargo-nextest` is not installed here

### DIFF
--- a/assets/macos/Kaku.app/Contents/Resources/kaku.lua
+++ b/assets/macos/Kaku.app/Contents/Resources/kaku.lua
@@ -104,7 +104,6 @@ local user_has_custom_padding = false
 local user_has_custom_font = false
 local user_has_custom_font_rules = false
 local remember_last_cwd = true
-local tab_title_show_basename_only = false
 
 local function check_user_custom_config()
   local user_config_path = kaku_user_config_path()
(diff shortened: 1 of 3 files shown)


rookie-ricardo:
Addressed in `8248af4`.

Changes in this update:
- moved `tab_title_show_basename_only` into the real `Config` schema so it works with `KAKU_STRICT_CONFIG=1`
- stopped text-scanning the user config file for this flag and now read it from `effective_config` inside `format-tab-title`
- shortened the settings label from `Basename-only Tab Titles` to `Short Tab Titles` so it fits the TUI layout better

tw93:
@rookie-ricardo Thanks for the contribution. I merged this and expect to ship it in an upcoming release soon. This is a nice usability improvement for tab titles.

PR #226, opened 2026-03-22
fix(gui): scroll viewport on trackpad wheel in alt screen

### LINKED ISSUE #225: macOS trackpad two-finger scroll triggers input history instead of scrolling content
## Description

When running Claude Code CLI (or any TUI app using the alternate screen) inside Kaku on macOS, two-finger trackpad scrolling navigates input history (as if pressing Up/Down arrow keys) instead of scrolling the conversation content.

The same CLI works correctly in iTerm2 — trackpad scrolling scrolls the viewport.

## Steps to Reproduce

1. Open Kaku on macOS
2. Run any alt-screen TUI app (e.g. Claude Code CLI, `htop`)
3. Use two-finger trackpad scroll

## Expected Behavior

Trackpad scrolling should scroll the viewport content (entering Primary Screen Peek if in alt screen).

## Actual Behavior

Trackpad scrolling triggers Up/Down arrow key events, causing input history navigation or list selection changes.

## Root Cause

In `kaku-gui/src/termwindow/mouseevent.rs`, when the alt screen is active and the app hasn't enabled mouse reporting, `bypass_wheel_assignment_in_alt` is set to `true`. This skips the mouse binding lookup (which would trigger `ScrollByCurrentEventWheelDelta`), and the wheel event falls through to `pane.mouse_event()`. The terminal emulator layer (`term/src/terminalstate/mouse.rs:121-134`) then converts wheel events to arrow keys — the xterm `alternateScroll` behavior.

## Environment

- macOS (Darwin 24.6.0)
- Kaku (latest main branch)

### LINKED ISSUE #227: Trackpad scroll past top of scrollback causes infinite loop and duplicate content
## Description

When using trackpad (two-finger) scrolling in alt screen mode (e.g., inside a TUI app), scrolling up past the top of the scrollback buffer causes the viewport to jump back to the bottom, creating an infinite scroll loop with duplicated content. iTerm2 does not have this issue — it stops at the top of scrollback.

## Steps to Reproduce

1. Open Kaku and run a TUI application (e.g., `claude` CLI, `htop`, etc.)
2. Use two-finger trackpad scroll to scroll up
3. Keep scrolling up past the very top of the scrollback history

## Expected Behavior

Scrolling should stop at the top of the scrollback buffer (matching iTerm2 behavior).

## Actual Behavior

The viewport jumps from the top of scrollback back to the bottom, and the content appears duplicated. There is no end to scrolling — it loops infinitely.

## Root Cause

In `scroll_by_line`, when the computed scroll position goes below `scrollback_top` (e.g., position becomes -1 while `scrollback_top` is 0), `normalize_viewport` treats this as "pruned scrollback" and snaps the viewport to the bottom. This creates a cycle: scroll up → hit top → snap to bottom → scroll up → same content again.

The fix is to clamp the position to `scrollback_top` before passing it to `set_viewport`, so the viewport stops at the top of scrollback instead of wrapping around.

guohai:
## Summary

- Fix macOS trackpad two-finger scroll triggering input history navigation (arrow keys) instead of scrolling content when running TUI apps (e.g. Claude Code CLI) in the alternate screen
- Route wheel events to `scroll_by_current_event_wheel_delta()` when `bypass_wheel_assignment_in_alt` is true, entering Primary Screen Peek mode — matching iTerm2 behavior
- Clamp scroll position to `scrollback_top` so scrolling stops at the top of the scrollback buffer instead of looping infinitely with duplicated content

Fixes #225
Fixes #227

## Root Cause

**Arrow key issue (#225):** In `mouseevent.rs`, when `bypass_wheel_assignment_in_alt` is `true` (alt screen active, no mouse grab), wheel events fell through to the terminal emulator which converted them to arrow keys via xterm's `alternateScroll` behavior.

**Infinite scroll loop (#227):** In `scroll_by_line`, when the scroll position went below `scrollback_top`, `normalize_viewport` interpreted it as "pruned scrollback" and snapped to the bottom, creating an infinite loop with duplicated content.

## Changes

1. Early return in `mouseevent.rs` to route wheel events to `scroll_by_current_event_wheel_delta()` instead of falling through
2. Added `.max(dims.scrollback_top)` in `scroll_by_line` to clamp the position at the top of scrollback

## Test plan

- [ ] Build and run Kaku on macOS
- [ ] Open Claude Code CLI (or any alt-screen TUI like `htop`)
- [ ] Two-finger trackpad scroll should scroll the viewport (Primary Screen Peek), not navigate input history
- [ ] Scrolling up past the top of scrollback stops at the top instead of looping
- [ ] Keyboard Up/Down still navigates history as expected
- [ ] Mouse-reporting apps (vim with mouse enabled) still handle scroll correctly
- [ ] `less` without alt screen still receives arrow keys from wheel

Generated with SMT <smt@agora.build>

### DIFF
--- a/kaku-gui/src/termwindow/mod.rs
+++ b/kaku-gui/src/termwindow/mod.rs
@@ -3597,7 +3597,8 @@ impl TermWindow {
         let position = self
             .effective_viewport(pane)
             .unwrap_or(dims.physical_top)
-            .saturating_add(amount);
+            .saturating_add(amount)
+            .max(dims.scrollback_top);
 
         self.reveal_scrollbar();
         self.set_viewport(pane.pane_id(), Some(position), dims);


tw93:
Thanks for the great PR, @guohai! Both fixes are clean and well-targeted. Merged — will be included in the next release in the coming days.

PR #232, opened 2026-03-23
fix(macos): normalize Finder service file targets to parent directories

### LINKED ISSUE #231: [BUG] Finder "New Kaku Tab Here / New Kaku Window Here" passes file paths instead of parent directories
## Summary

On macOS Finder, when using `New Kaku Tab Here` or `New Kaku Window Here` on a regular file, Kaku forwards the file path itself instead of using the file's parent directory.

This causes downstream logic to treat a file as a working directory and makes the action fail.

## Steps to reproduce

1. In Finder, locate any regular file such as `README.md`
2. Right-click the file and choose `New Kaku Tab Here`
3. Or right-click the file and choose `New Kaku Window Here`

## Expected vs actual

- **Expected**: Kaku opens a new tab or window in the file's parent directory
- **Actual**: Kaku forwards the file path itself, which is later treated as a directory and causes the action to fail

## Environment

Please include:
- **Kaku version**: `0.8.0`
- **macOS version**: `15.7.2 (24G325)`
- **Machine**: `MacBook Air M1 2020`
- **Shell**: `zsh`

## Optional details

Add anything helpful if available:
- The issue appears to be limited to the Finder service entry path
- Using the same actions on a directory behaves as expected
- I can provide a screen recording or related logs if needed

## Extra context

- The root cause is that the Finder service entry point does not distinguish between file targets and directory targets
- For file targets, the correct behavior is to use the parent directory rather than the file path itself

LanternCX:
## Summary

- Fix Finder `New Kaku Tab Here` / `New Kaku Window Here` forwarding file paths directly into the service open flow
- Normalize explicit file targets to their parent directories while leaving directory targets unchanged
- Keep the change scoped to the Finder service path only, without changing other open-file or URL entry points

Closes #231 

## Root Cause

The macOS Finder service entry points accepted the first incoming path and forwarded it as-is.

That behavior worked for directory targets, but when Finder invoked the service on a file, Kaku propagated the file path itself. Downstream logic then treated that value as a working directory, which caused `not a dir` style failures.

## Changes

1. Add a small path normalization step in the macOS Finder service flow
2. Convert explicit file targets to their parent directories
3. Preserve existing behavior for directory targets and unknown paths
4. Add focused regression tests for:
   - file target -> parent directory
   - directory target -> unchanged
   - unknown path -> unchanged

## Test plan

- [x] `cargo test -p window normalize_finder_service_path -- --nocapture`
- [x] `cargo build -p kaku-gui`
- [x] `make app`
- [x] Manual verification: right-clicking a file in Finder with `New Kaku Tab Here` opens Kaku in the file's parent directory
- [x] Manual verification: right-clicking a file in Finder with `New Kaku Window Here` opens Kaku in the file's parent directory
- [x] Manual verification: right-clicking a directory still behaves the same as before

## Notes

- Local verification also required restoring missing vendored dependency working trees before the build could run successfully

### DIFF
--- a/window/src/os/macos/app.rs
+++ b/window/src/os/macos/app.rs
@@ -21,6 +21,7 @@ use objc::*;
 use std::cell::RefCell;
 use std::convert::TryFrom;
 use std::ffi::c_void;
+use std::path::Path;
 use std::process::Command;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Mutex;
@@ -802,6 +803,17 @@ extern "C" fn keyboard_selection_did_change(
 #[allow(clippy::items_after_test_module)]
 mod tests {
     use super::*;
+    use std::fs;
+    use std::path::PathBuf;
+    use std::time::{SystemTime, UNIX_EPOCH};
+
+    fn unique_test_path(name: &str) -> PathBuf {
+        let nanos = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("system time before unix epoch")
+            .as_nanos();
+        std::env::temp_dir().join(format!("kaku-app-tests-{}-{nanos}", name))
+    }
 
     #[test]
     fn layout_translation_modifier_flags_include_shift_and_option() {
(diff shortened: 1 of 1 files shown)


tw93:
@LanternCX The change is correctly scoped to the Finder service path, preserves the existing behavior for other open-file and URL entry points, and adds focused regression coverage for file, directory, and unknown-path inputs.

I also reran make fmt-check, make check, make test, and make app locally. Thanks.

PR #234, opened 2026-03-23
Add Restore Previous Window for multi-tab window snapshots

rookie-ricardo:
## Summary

Add a `Window > Restore Previous Window` action that restores the most recently saved multi-tab window snapshot.

This is intentionally scoped to restore window structure only:
- window tabs
- active tab
- pane split layout inside each tab
- each pane's working directory
- window and tab titles

It does not attempt to restore the original running shell processes.

## Problem

Kaku currently remembers window geometry, but it does not preserve a useful workspace shape across app restarts or app updates.

That is especially painful during update/restart flows:
- users may have one important window with multiple tabs open
- after restart, all tabs are gone
- rebuilding that layout manually is repetitive, even if restoring the original shell processes is out of scope

For this use case, restoring the last meaningful window layout is already a big improvement.

## What this PR adds

This PR adds a lightweight snapshot-and-restore flow:

- keep a single snapshot file at `~/.config/kaku/last_window_session.json`
- only snapshot the currently focused window when it has more than one tab
- overwrite the previous snapshot with the latest eligible one
- add `Window > Restore Previous Window`
- restore that snapshot into a new window on demand

The snapshot is not deleted automatically after restore.
It stays available until a newer eligible window snapshot replaces it.

### Save behavior

A snapshot is written for the currently focused window when:
- Kaku exits normally
- relevant mux/window changes happen and the focused window is eligible

A window is eligible only when:
- it exists
- it is the focused window
- it has more than one tab

Single-tab windows do not replace the saved snapshot.

### Restore behavior

`Window > Restore Previous Window`:
- reads `last_window_session.json`
- creates a new window in the current workspace
- recreates each saved tab
- recreates each tab's pane split tree
- spawns fresh panes using the saved working directories
- restores active tab selection
- restores saved tab/window titles when present

This restores layout and navigation context, not live process state.

## Why this design

This PR intentionally does not implement full session persistence.

A full session restore would require keeping shell processes alive across app restarts, which is a much larger architecture change.

This feature instead targets the most practical recovery case:
- "I restarted or updated Kaku"
- "I want my previous multi-tab window shape back"
- "I do not need the original running processes to survive"

That keeps the implementation small, predictable, and easy to reason about.

## Notes

- hard crashes / `kill -9` / power loss are still best-effort only
- restore uses the latest saved eligible snapshot
- the snapshot is intentionally persistent and only replaced by a newer one

## Validation

I validated the feature with:

- `cargo fmt --all`
- `cargo check -p kaku-gui -p config`

And manually tested the flow:

1. open a window with 2+ tabs
2. quit Kaku
3. relaunch Kaku
4. use `Window > Restore Previous Window`
5. confirm tabs/layout/cwds are recreated in a new window

### DIFF
--- a/kaku-gui/src/frontend.rs
+++ b/kaku-gui/src/frontend.rs
@@ -343,15 +343,8 @@ impl GuiFrontEnd {
                     })
                     .detach();
                 }
-                MuxNotification::TabTitleChanged { tab_id, .. } => {
-                    let mux = Mux::get();
-                    if let Some(window_id) = mux.window_containing_tab(tab_id) {
-                        crate::session_restore::request_save_window_snapshot(window_id);
-                    }
-                }
-                MuxNotification::WindowTitleChanged { window_id, .. } => {
-                    crate::session_restore::request_save_window_snapshot(window_id);
-                }
+                MuxNotification::TabTitleChanged { .. } => {}
+                MuxNotification::WindowTitleChanged { .. } => {}
                 MuxNotification::TabResized(tab_id) => {
                     let mux = Mux::get();
                     if let Some(window_id) = mux.window_containing_tab(tab_id) {
@@ -362,9 +355,7 @@ impl GuiFrontEnd {
                     crate::session_restore::request_save_window_snapshot(window_id);
                 }
                 MuxNotification::PaneRemoved(_) => {}
-                MuxNotification::WindowInvalidated(window_id) => {
-                    crate::session_restore::request_save_window_snapshot(window_id);
-                }
+                MuxNotification::WindowInvalidated(_) => {}
                 MuxNotification::PaneOutput(_) => {}
                 MuxNotification::PaneAdded(pane_id) => {
                     let mux = Mux::get();


rookie-ricardo:
这是一个意向 PR，我不确定我这种设计是否符合 Kaku 的未来规划。

rookie-ricardo:
Addressed the review comments in `ec97430`:

- save requests now target the specific window/tab that changed instead of a delayed `focused_window_id()` lookup
- snapshots now persist each pane's original domain name and restore against that domain instead of always using `mux.default_domain()`
- restore now cancels the provisional window if reconstruction fails, so it doesn't leave a partial empty window behind

I also added an explicit save before closing the current GUI window so the "close one multi-tab window while another window stays focused" case does not snapshot the wrong survivor window.

tw93:
Hey @rookie-ricardo, thanks for the well-structured PR! I made a small tweak, removed the snapshot triggers on `WindowInvalidated`, `TabTitleChanged`, and `WindowTitleChanged` since those fire very frequently and would cause continuous disk writes. Now snapshots only happen on structural changes (tab added, pane added, resize) and reliably on window close and exit.

That said, I'm going to hold off on merging for now. The feature itself is thoughtfully scoped, but session restore isn't something most Kaku users are asking for yet. I'll revisit when there's more community demand — if this becomes a frequently requested feature, it'll be a clear signal to ship it. Appreciate the effort!

PR #250, opened 2026-03-28
fix(gui): add always-on-top to window menu

FelixLyfe:
add the existing Always on Top action to the default Window menu command list

### DIFF
--- a/kaku-gui/src/commands.rs
+++ b/kaku-gui/src/commands.rs
@@ -740,6 +740,7 @@ impl CommandDef {
                 "Window" => match action {
                     Hide => 10,
                     ToggleFullScreen => 12,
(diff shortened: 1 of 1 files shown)


tw93:
Hi @FelixLyfe, clean and minimal change. `ToggleAlwaysOnTop` already had a keyboard shortcut but wasn't discoverable via the Window menu，this makes it visible. The rank 13 placement after `ToggleFullScreen` makes sense.

tw93:
After thinking it through, I decided to remove **Always on Bottom** from the Window menu and keyboard shortcut. For a terminal, keeping a window behind everything else isn't a meaningful use case,users who need fine-grained window level control can still use `SetWindowLevel` via Lua config. **Always on Top** stays as-is. Thanks for the contribution\!

PR #254, opened 2026-03-30
feat(config): add macOS traffic lights setting

emrezeytin:
## Summary
- add a dedicated macOS Traffic Lights setting in the Settings UI so users can hide the integrated traffic lights without editing Lua by hand
- keep Shadow as a separate toggle while still persisting a single `window_decorations` config value
- preserve unsupported `window_decorations` values unless the user explicitly edits the setting
- document the `window_decorations = \"RESIZE\"` override and add regression coverage for config load/save behavior

## Testing
- make fmt
- make check
- make test

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -99,7 +99,7 @@ By default, Kaku embeds the macOS traffic light buttons into the tab bar area us
 config.window_decorations = "RESIZE"
 ```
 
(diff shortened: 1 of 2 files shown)


tw93:
@emrezeytin thanks, I merged this. Before landing it, I added a small follow-up so the config TUI preserves valid `INTEGRATED_BUTTONS` variants, including the no-`RESIZE` forms, instead of normalizing them when only the shadow setting changes. I also added regression coverage for those cases and cleaned up the related docs wording.

PR #262, opened 2026-04-04
feat(selection): add select-all shortcut for visible pane content

### LINKED ISSUE #261: [FEATURE] Add a shortcut to select the current window content
Add a shortcut to select the current window content

## Problem

Kaku currently supports mouse-based selection for copying terminal output, but there is no keyboard shortcut that directly selects all visible content in the current pane or window.

On macOS, users often expect common selection actions to also be available from the keyboard.

The current meaning of `Cmd+A` is closer to selecting the whole command line, but similar line-selection behavior can already be achieved with `Cmd+Shift+Left`. What is still missing is a shortcut for selecting the current visible region.

## Proposed behavior

Kaku should support a dedicated shortcut for selecting all visible content in the active pane.

One possible direction is:

- use `Cmd+A` on macOS to select the visible content of the current pane
- give that shortcut a single clear meaning instead of mixing multiple behaviors

That said, whether `Cmd+A` is the right shortcut, or whether another shortcut would be better, is still open for discussion.

## Use case

A common scenario is that a user runs a command in the terminal and wants to quickly copy the output currently visible on screen.

Today this usually requires mouse selection. With a keyboard shortcut, the flow becomes much smoother: select the visible content, then copy it.

## Optional references

- Whether `Cmd+A` should be repurposed to select the whole visible page should be considered in the context of the existing design direction
- An alternative that was discussed is keeping the current meaning of `Cmd+A` and using a double-press `Cmd+A` flow to expand from line selection to page selection
- I already prepared a draft implementation that maps `Cmd+A` to selecting all visible content in the current pane, which can serve as a discussion starting point

## Checklist

- [x] I searched existing issues and this hasn't been reported yet

LanternCX:
## Summary

- This PR provides one possible implementation direction for adding keyboard-based "select all" support for visible pane content.
- In this approach, `Cmd+A` on macOS is changed to select **all visible content** in the current pane.

## What Problem This Tries To Solve

Kaku currently supports mouse-based selection, but there is no keyboard shortcut that directly selects all visible content in the current terminal window.

This makes common actions like quickly copying terminal output depend on the mouse, which is less convenient.

Close #261 

## Current Approach

- Add a GUI-level `SelectAll` action.
- Bind `Cmd+A` to that action in the default macOS config.
- Select the full visible region of the active pane.
- Remove the previous `Cmd+A` behavior that selected only the current command line, so the shortcut has a single clear meaning.

## Alternatives Considered

- Keep the current behavior and add another shortcut for full-window selection.
- Use a two-step approach: the first `Cmd+A` keeps the current behavior, and the second expands selection to the visible region.

There are multiple possible solutions, and they are also discussed in the #261 . Because of that, this PR should still be treated as a Draft pending further discussion.

## Verification

- [x] I manually tested the packaged app, and the current approach behaves as expected.


LanternCX:
CI failure is from config lua::test::bundled_kaku_lua_compiles.
This appears unrelated to the selection shortcut changes in this PR, It looks more like an existing issue on the current base / merge result

PR #280, opened 2026-04-13
fix(macos): prevent use-after-free crash during window close with fullscreen transition

yxspace:
## Problem

When closing a window while it is in a fullscreen transition (entering or exiting), AppKit continues to fire transition callbacks (`will/did_enter/exit_fullscreen`) after the window has already been deallocated, causing a use-after-free crash.

## Fix

- Add an `is_closing: Cell<bool>` flag to `WindowView`
- Set the flag to `true` at the start of `window_will_close`, before any cleanup
- Guard all fullscreen transition handlers with an early return when `is_closing` is set
- Replace `borrow_mut()` with `try_borrow_mut()` in transition handlers to avoid `BorrowMutError` panics when AppKit fires notifications synchronously while `inner` is already borrowed during close

## Testing

Reproduce: open a window, enter fullscreen, then immediately close the window before the transition completes. Without this fix, the app crashes. With this fix, the transition callbacks are safely ignored.

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -2763,6 +2763,9 @@ struct WindowView {
     native_fullscreen_target: Cell<Option<bool>>,
     native_fullscreen_transition_start: Cell<Option<Instant>>,
     resize_retry_scheduled: Cell<bool>,
+    /// Set when window_will_close fires; prevents fullscreen transition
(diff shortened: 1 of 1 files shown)


tw93:
@yxspace thanks for fixing the use-after-free crash during fullscreen window close, merged and will be in the next release.

PR #282, opened 2026-04-13
fix(clipboard): avoid clearing clipboard on empty cmd-c

beautifulrem:
## Summary

  Fix a clipboard regression when `copy_on_select` is enabled.

  Previously, selecting text with the mouse correctly copied the selection to the clipboard, but pressing `Cmd+C` on the same
  selection could overwrite the clipboard with an empty string. This broke the expected idempotent behavior of copy.

## Root Cause

  The mouse-selection completion path already guarded against empty selection text before writing to the clipboard.

  The `Cmd+C` path (`CopyTo`) did not. It recomputed `selection_text()` and always wrote the result to the macOS pasteboard.
  If that recomputation returned an empty string, the clipboard was cleared.

## Changes

  - Add a guarded clipboard helper that only writes when the text is non-empty.
  - Route the `Cmd+C` / `CopyTo` path through that helper.
  - Keep explicit text-copy paths unchanged.
  - Add regression tests for the empty/non-empty clipboard text guard.

### DIFF
--- a/kaku-gui/src/termwindow/clipboard.rs
+++ b/kaku-gui/src/termwindow/clipboard.rs
@@ -40,6 +40,18 @@ fn should_emit_ai_notice(kind: &str, message: &str) -> bool {
 }
 
 impl TermWindow {
+    pub fn copy_to_clipboard_if_present(
+        &self,
+        clipboard: ClipboardCopyDestination,
+        text: String,
+    ) -> bool {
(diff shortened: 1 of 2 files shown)


tw93:
@beautifulrem thanks for fixing the clipboard clear-on-empty-copy bug, merged and will be in the next release.

PR #290, opened 2026-04-18
fix(config): exclude trailing punctuation and wrappers from file link regex

ddotz:
## Summary

The implicit file-path hyperlink rule matched one character too far past the end of the actual path, so clicking a detected link was often a silent no-op.

**Cases the current regex fails on:**

- Sentence-final punctuation: `see docs/foo.md.`, `see docs/foo.md,`, `open docs/foo.md?`
- CJK particles attached without whitespace — very common in Korean prose emitted by AI CLI tools:
  - `docs/foo.md에 저장했습니다` (locative)
  - `docs/foo.md를 열어주세요` (object)
  - `docs/foo.md은 업데이트됨` (topic)
- Matched wrapper characters: `` `docs/foo.md` ``, `'docs/foo.md'`, `"docs/foo.md"`

In each case the rule captures `docs/foo.md.` / `docs/foo.md에` / `` docs/foo.md` `` and builds a `file://...` URI that doesn't exist on disk. The click handler's `target.path.exists()` check fails, so the click does nothing even though the path is visibly underlined.

## Fix

Change the trailing portion of the rule so the last captured character must be an ASCII word character, `/`, or `-`:

```diff
- r"(^|[\s\(\[<])((?:~|\.{1,2}|[[:alnum:]_.-]+)?/[^\s\)\]\}>]+)"
+ r#"(^|[\s\(\[<`'"])((?:~|\.{1,2}|[[:alnum:]_.-]+)?/[^\s\)\]\}>`'"]*[a-zA-Z0-9_/-])"#
```

Regex backtracking naturally drops trailing characters that don't fit, so `docs/foo.md에` backtracks to `docs/foo.md`. Backticks and straight quotes are added to the leading boundary class and excluded from the inner class so wrapped paths are detected with the wrappers stripped.

**The rule remains extension-neutral** — `.rs`, `.py`, `.ts`, `.go`, `.json`, extensionless files like `Makefile`/`Dockerfile`, and dotfile-style `.env.local` all continue to match because they end in an ASCII letter or digit. `file:line` and `file:line:col` suffixes are preserved for the same reason (trailing digit).

## Test plan

- [x] `cargo test -p config --lib` — 36 passed, 0 failed
- Existing tests untouched: `file_hyperlink_rule_matches_bare_relative_paths`, `file_hyperlink_rule_does_not_override_urls`
- New tests:
  - `file_hyperlink_trims_trailing_ascii_punctuation` — `.`, `,`, `;`, `:`, `!`, `?`
  - `file_hyperlink_trims_trailing_korean_particles` — `에`, `를`, `은`
  - `file_hyperlink_handles_wrapping_delimiters` — backtick, single quote, double quote, parentheses
  - `file_hyperlink_preserves_line_and_column` — `:42`, `:42:10`, and `:42:10.` regressions
  - `file_hyperlink_handles_absolute_paths_with_particles` — the motivating bug: `/Users/x/Code/proj/docs/plans/2026-04-18-v8-plan.md` followed by `에`
  - `file_hyperlink_is_extension_neutral` — pins the invariant across `.rs`, `.py`, `.ts`, `.go`, `.json`, `.env.local`, `Makefile`, `Dockerfile`

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -2109,8 +2109,16 @@ pub fn default_hyperlink_rules() -> Vec<hyperlink::Rule> {
         // File paths: support absolute paths, common relative prefixes, and
         // bare relative paths like `kaku/src/main.rs`.
         // Supports file:line and file:line:col formats.
+        //
+        // The trailing character class is intentionally restricted to ASCII
+        // word/`/`/`-` so that trailing punctuation (`.`, `,`, `;`, `:`, `!`,
+        // `?`), CJK particles (e.g. Korean `에`/`을`), or matched wrappers
+        // (backticks, quotes) attached after a path are excluded from the
+        // match via regex backtracking. The leading boundary class likewise
+        // admits backticks and straight quotes so a path wrapped in those
+        // characters is still detected.
         hyperlink::Rule::with_highlight(
-            r"(^|[\s\(\[<])((?:~|\.{1,2}|[[:alnum:]_.-]+)?/[^\s\)\]\}>]+)",
+            r#"(^|[\s\(\[<`'"])((?:~|\.{1,2}|[[:alnum:]_.-]+)?/[^\s\)\]\}>`'"]*[a-zA-Z0-9_/-])"#,
             "file://$2",
             2,
         )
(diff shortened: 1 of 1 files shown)


ddotz:
Heads up on the red `check_release_config.sh` check — this failure exists on `main` already and is unrelated to this change:

- This PR only touches `config/src/config.rs`; it does not modify `assets/shell-integration/config_update_highlights.tsv`.
- On `main`, `grep '^20' assets/shell-integration/config_update_highlights.tsv` returns **one entry** (English only):
  > `20 Zsh comment color is now visible in dark themes; no longer invisible against dark backgrounds`
- `scripts/check_release_config.sh` requires **at least 2** entries per version (English + Chinese), so any PR targeting `main` currently fails this check regardless of content.

Happy to rebase once the missing Chinese highlight entry for version 20 is added on `main`, or to include it here if you'd prefer — just didn't want to silently bundle unrelated translation copy into a regex-fix PR.

PR #294, opened 2026-04-19
fix: reset viewport on window resize to prevent content jump and loss

CoderGoGoGo:
## Summary
- Reset all pane viewports to bottom after every window resize in `apply_dimensions()`, preventing content from appearing to jump or becoming inaccessible
- Limit pre-prune blank line removal in `Screen::resize()` to avoid cumulative content loss during repeated resizes

## Problem
When resizing the terminal window:
1. The last line of content jumps upward
2. Cannot scroll up to see previous content
3. Repeated resizes progressively clear terminal content

## Root Cause
`physical_top` shifts on resize because both `lines.len()` and `physical_rows` change, but `PaneState::viewport` was never adjusted. Additionally, the pre-prune loop in `Screen::resize()` had no upper bound, causing cumulative content loss.

## Test Plan
- [ ] Output several screenfuls of content, then resize the window — content should not jump
- [ ] After resize, scroll up — all previous content should be accessible
- [ ] Rapidly resize multiple times — content should not be cleared

### DIFF
--- a/kaku-gui/src/termwindow/resize.rs
+++ b/kaku-gui/src/termwindow/resize.rs
@@ -555,6 +555,12 @@ impl super::TermWindow {
                     tab.resize(size);
                 }
             }
+            for tab in window.iter() {
+                for pos in tab.iter_panes_ignoring_zoom() {
+                    let mut state = self.pane_state(pos.pane.pane_id());
+                    state.viewport = None;
+                }
+            }
         };
         if live {
             self.pending_pty_flush_after_resize = true;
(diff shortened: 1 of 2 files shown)


tw93:
@CoderGoGoGo 感谢 PR，问题定位是对的。调整了一下 viewport 的处理方式：resize 后对已有 viewport 做 normalize 而不是直接重置为 None，这样用户在查看历史内容时 resize 不会被拉回底部。pre-prune 的上限逻辑保留了你的思路。已合入 main。

PR #314, opened 2026-04-24
perf(mux): eliminate per-subscriber MuxNotification clones in hot path

darion-yaphet:
## Summary

   - Change `Mux::subscribe()` to accept `Fn(&MuxNotification) -> bool` instead of `Fn(MuxNotification) -> bool`
   - `notify()` now passes `&notification` to each subscriber, removing the per-subscriber `.clone()` call
   - Subscribers that need ownership (dispatch channel, client domain, tmux commands, termwindow spawn) clone explicitly at their call site; read-only subscribers
   dereference `PaneId` directly

   ## Motivation

   `MuxNotification::PaneOutput` is emitted thousands of times per second during active terminal output. With the previous design, every notification was cloned once
    per registered subscriber — O(N × subscribers) heap allocations — even though most subscribers only read the `PaneId` and return early.

   Notable `MuxNotification` variants that carry heap-allocated data:
   - `TabTitleChanged { title: String }`
   - `WindowTitleChanged { title: String }`
   - `WorkspaceRenamed { old_workspace: String, new_workspace: String }`

   These clones were unconditional regardless of whether the subscriber ever used the value.

   ## Changes

   | File | Change |
   |------|--------|
   | `mux/src/lib.rs` | Subscriber type → `Fn(&MuxNotification) -> bool`; `notify()` passes `&notification` |
   | `mux/src/tmux_commands.rs` | `let n = n.clone()` before `spawn_into_main_thread` |

  - [ ] Build `mux`, `kaku-gui`, `crates/kaku-remote`, `crates/wezterm-client`, `crates/wezterm-mux-server-impl` successfully
  - [ ] Verify tmux integration still handles `PaneFocused` notifications

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,4 @@ website/dist/
 website/.astro/
 website/test-results/
 website/playwright-report/
+.worktrees/
--- a/crates/kaku-remote/src/lib.rs
+++ b/crates/kaku-remote/src/lib.rs
@@ -805,7 +805,7 @@ pub fn start_tunnel(tunnel_url: String) {
         let tunnel_throttle: PaneThrottle = Arc::new(Mutex::new(HashMap::new()));
         mux.subscribe(move |notification| {
             if let MuxNotification::PaneOutput(pane_id) = notification {
-                let pid: usize = pane_id.into();
+                let pid: usize = (*pane_id).into();
                 if tx_for_sub.receiver_count() > 0 {
                     {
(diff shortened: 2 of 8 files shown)


tw93:
@darion-yaphet thanks for the optimization, the direction is right. I cherry-picked the core change (pass `&MuxNotification` instead of cloning per subscriber), dropped the unrelated `.gitignore` commit, rebased onto main and merged.

843 tests pass, will ship in the next release.

PR #324, opened 2026-04-30
Add configurable file link open targets

s010s:
## Summary

This adds a macOS-only default open target setting for Cmd-clicked file links, so users are no longer forced through the previous VS Code-oriented behavior.

The new setting is exposed through the existing `kaku config` TUI under an Integrations section. Kaku detects available editors and terminal apps when the TUI opens, shows only targets that are actually present, and keeps a previously configured missing target visible as a disabled value so the config state remains understandable.

Supported targets include:

- Auto
- Default app, Finder, Terminal
- VS Code, Cursor, Windsurf, Kiro, Antigravity, Zed, IntelliJ IDEA
- iTerm2, Ghostty, WezTerm, cmux

## Details

- Added a typed `default_open_target` config field with Lua string serialization and parsing.
- Added lightweight macOS target detection for PATH commands, common app bundles, and known bundle-contained CLI paths.
- Moved GUI file-link opening into a dedicated command-building module so behavior is testable without launching apps.
- Preserved the existing path resolution behavior for `file`, `file:line`, and `file:line:col` links.
- Added target-specific command construction:
  - VS Code, Cursor, Windsurf, and Antigravity use goto-style file locations where available.
  - Zed uses suffix-style file locations.
  - IntelliJ IDEA uses `--line` and `--column` flags.
  - Terminal-style targets open the containing directory for file links and the directory itself for directory links.
  - Finder reveals files and opens directories.
  - Default app uses the macOS default handler.
- Added Auto fallback ordering that tries `$VISUAL` / `$EDITOR` from the GUI process environment first, then detected IDE targets, then macOS default app, then Finder.
- Kept opening failures silent in the UI, with fallback continuing through the remaining candidates.

## Test Plan

- `cargo test -p config default_open_target`
- `cargo test -p kaku open_target`
- `cargo test -p kaku-gui open_target`
- `cargo fmt --all -- --check`
- `git diff --cached --check`

Manual checks covered selecting detected targets in the config TUI and Cmd-clicking file links with line/column suffixes for editor targets and directory-opening behavior for terminal targets.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -222,6 +222,9 @@ pub struct Config {
     #[dynamic(default = "default_gui_startup_args")]
     pub default_gui_startup_args: Vec<String>,
 
+    #[dynamic(default)]
+    pub default_open_target: DefaultOpenTarget,
+
     /// Specifies the default current working directory if none is specified
     /// through configuration or OSC 7 (see docs for `default_cwd` for more
     /// info!)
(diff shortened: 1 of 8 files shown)


tw93:
@s010s thanks for putting this together. Kaku tries to stay editor-neutral, so adding built-in presets for specific IDEs isn't something we'd merge. The existing `$EDITOR`/`$VISUAL` env var path is the intended way to configure this. Closing for now, but appreciate the effort.

s010s:
> editor

I understand the editor-neutral concern, and I agree that Kaku should not force users toward a specific editor.

The reason I opened this PR is that the current behavior is already not fully editor-neutral in practice: Cmd-clicked file links try VS Code-specific handling before falling back to `$VISUAL` / `$EDITOR` and then macOS default open behavior. For users of Cursor, Zed, Windsurf, terminal-first workflows, or simply macOS default apps, there is currently no discoverable setting in `kaku config` to change that behavior.

I also could not find user-facing documentation explaining that `$EDITOR` / `$VISUAL` is the intended way to configure Cmd-click file links. Since Kaku is a macOS GUI app, users who launch it from Finder/Dock may also not naturally get their shell environment in the GUI process, which makes this path hard to discover and hard to rely on.

That said, I understand that the named IDE presets in this PR may be too product-specific for Kaku. Would you be open to a smaller editor-neutral version of this change?

For example:

- keep `Default app`, `Finder`, and `Terminal` as neutral macOS targets;
- keep `$VISUAL` / `$EDITOR` as the first-class configurable path;
- remove the built-in IDE presets from the settings UI;
- add documentation explaining how Cmd-click file links are opened and how `$EDITOR` / `$VISUAL` should be configured.

My goal is not to make Kaku favor any specific IDE, but to make the current file-link opening behavior explicit, discoverable, and configurable for users who do not use VS Code.

PR #331, opened 2026-05-04
chore: gitignore .gitnexus index dir

2233admin:
Why: gitnexus indexer drops state into .gitnexus/ at repo root; not source.


tw93:
Thanks! This is already covered on main by 16a6399 (chore: ignore gitnexus index), so I'm closing this duplicate before the 0.10.0 release.

PR #361, opened 2026-05-11
feat(mouse): allow scrollback wheel + selection extend during drag

MingmouHaochi:
## Why

When dragging a terminal selection with the left mouse button held
down, the wheel event is currently dropped. This forces users to
release the mouse, scroll, then re-select to grab text spanning more
than one screen of output — a common case in log inspection and stack
traces.

Every macOS `NSTextView` app — Safari, TextEdit, Notes, VS Code,
Sublime, iTerm2, and `Terminal.app` itself — scrolls **and** extends
the selection when the wheel fires during a drag. Kaku is currently
the odd one out.

## What

Adds a new config option `selection_wheel_scroll_behavior` with three
modes:

| Mode | Behavior |
| :--- | :--- |
| `"Extend"` (default) | scroll viewport **and** extend selection to follow the cursor under the new viewport |
| `"ScrollOnly"` | scroll viewport, leave the selection range unchanged |
| `"Ignore"` | drop the wheel event entirely (the v0.10 legacy behavior) |

### Implementation notes

- `should_suppress_wheel_during_terminal_selection` is replaced with
  `wheel_during_terminal_selection_action`, which returns an action
  enum instead of a boolean.
- The wheel short-circuit in `mouse_event_impl` becomes a dispatcher;
  the actual scroll + selection update happens in a new
  `handle_wheel_during_terminal_selection` method.
- `ScrollAndExtend` recomputes the `StableRowIndex` of the mouse
  position against the **post-scroll** viewport, refreshes
  `pane_state.mouse_terminal_coords`, and then reuses the existing
  `extend_selection_at_mouse_cursor` helper — so all selection-mode
  semantics (`Cell` / `Word` / `Line` / `SemanticZone`) flow through
  the same code path.
- Alt-screen mouse reporting (vim / htop / tmux) is unaffected: the
  existing `should_bypass_wheel_assignment_in_alt` guard still wins
  over the new dispatcher for that path.

### Tests

The three legacy unit tests are replaced with seven covering each
behavior mode across the relevant call-site axes (terminal capture vs
UI vs none, left-button vs right-button, default vs explicit
behavior). All 18 `mouseevent::tests` and full `make check` /
`cargo nextest` pass on macOS 26 / aarch64.

## ⚠️ Default behavior change

Earlier Kaku versions behaved as if `"Ignore"` were set. Users who
relied on the legacy "wheel does nothing during drag" can restore it
via:

```lua
config.selection_wheel_scroll_behavior = "Ignore"
```

The docs include a `Default change in v0.11` callout pointing this
out.

## Test plan

- [ ] In any terminal, run `cat /etc/hosts` (or any multi-screen
      output)
- [ ] Hold left mouse button on first row, drag down to bottom of
      screen
- [ ] **Without releasing**, scroll the wheel down — viewport should
      scroll AND the selection should grow to follow the cursor
- [ ] Release mouse, verify clipboard (if `copy_on_select` is on)
      contains the full cross-screen text
- [ ] Set `selection_wheel_scroll_behavior = "Ignore"` in `kaku.lua`,
      restart Kaku, verify v0.10 behavior is restored
- [ ] Open `vim`, hold left mouse, scroll wheel — vim's own mouse
      handling is unchanged


tw93:
Cherry-picked as 0727745 on main. Thank you @MingmouHaochi for the clean implementation. The PR branch had stale commits from before V0.10.0, so cherry-pick was cleaner than merge.

Issue #305, opened 2026-04-21
[BUG] Pressing “⌘ + ,“ reduces the app's window in the Dock


julienducos:
## Summary

Pressing ⌘ + , does not open the app's settings but minimizes the app window to the Dock (French AZERTY keyboard).
This issue does not occur in other apps

## Steps to reproduce

1. Open the app
2. Press “⌘ + ,“
3. The app is reduced in the Dock

## Expected vs actual

- **Expected**: Open the app's settings
- **Actual**: Reduce the app in the Dock

## Environment

Please include:
- **Kaku version**: 0.9.0
- **macOS version**: macOS 26.4.1
- **Machine**: MacBoor Pro 14" M1 Pro
- **Shell**: zsh

## Optional details

Add anything helpful if available:
- screenshots or recordings
- relevant config snippet from `~/.config/kaku/kaku.lua`:
local wezterm = require 'wezterm'

local function resolve_bundled_config()
  local resource_dir = wezterm.executable_dir:gsub('MacOS/?$', 'Resources')
  local bundled = resource_dir .. '/kaku.lua'
  local f = io.open(bundled, 'r')
  if f then
    f:close()
    return bundled
  end

  local dev_bundled = wezterm.executable_dir .. '/../../assets/macos/Kaku.app/Contents/Resources/kaku.lua'
  f = io.open(dev_bundled, 'r')
  if f then
    f:close()
    return dev_bundled
  end

  local app_bundled = '/Applications/Kaku.app/Contents/Resources/kaku.lua'
  f = io.open(app_bundled, 'r')
  if f then
    f:close()
    return app_bundled
  end

  local home = os.getenv('HOME') or ''
  local home_bundled = home .. '/Applications/Kaku.app/Contents/Resources/kaku.lua'
  f = io.open(home_bundled, 'r')
  if f then
    f:close()
    return home_bundled
  end

  return nil
end

local config = {}
local bundled = resolve_bundled_config()

if bundled then
  local ok, loaded = pcall(dofile, bundled)
  if ok and type(loaded) == 'table' then
    config = loaded
  else
    wezterm.log_error('Kaku: failed to load bundled defaults from ' .. bundled)
  end
else
  wezterm.log_error('Kaku: bundled defaults not found')
end

-- User overrides:
-- Kaku intentionally keeps WezTerm-compatible Lua API names
-- for maximum compatibility, so `wezterm.*` here is expected.
-- Full API docs: https://wezfurlong.org/wezterm/config/lua/
[...]

Issue #220, opened 2026-03-20
[BUG] Kaku sometimes quits cleanly without a crash report


Sure-Will:
## Summary

Kaku sometimes disappears without a crash dialog, and the retained macOS logs suggest it is exiting through a normal quit path rather than crashing.

## Steps to reproduce

I do not have stable reproduction steps yet.

Observed twice so far:
- `2026-03-16 12:55:03 +0800`
- `2026-03-20 17:19:59 +0800`

In both cases, Kaku disappeared unexpectedly while I was using the machine. On the second occurrence, I manually reopened it later at around `2026-03-20 17:44 +0800`.

## Expected vs actual

- **Expected**: Kaku should remain open unless I explicitly quit it.
- **Actual**: Kaku sometimes disappears unexpectedly, but system logs show a clean quit path instead of a crash.

## Environment

- **Kaku version**: `0.7.1`
- **macOS version**: `26.3.1 (25D2128)`
- **Machine**: `Apple Silicon (arm64)`
- **Shell**: `zsh`

## Optional details

What I checked locally:
- My user config does not override quit/close behavior; it only loads bundled defaults.
- `/Applications/Kaku.app` was not updated on the incident day. Current app bundle mtime is `2026-03-14 22:14:17 +0800`.
- There is no Kaku crash report in `~/Library/Logs/DiagnosticReports` or `~/Library/Logs/CrashReporter`.

Relevant logs for the most recent incident (`2026-03-20 17:19:59 +0800`):
- `17:19:59.850` `WindowServer`: a `kCGSEventKeyDown` was still being delivered to Kaku (`pid 66786`)
- `17:19:59.937` `WindowServer`: `Process death ... (Kaku) ... pid: 66786`
- `17:19:59.937` `kaku-gui`: `Entering exit handler.`
- `17:19:59.957` `runningboardd`: `termination reported by launchd (0, 0, 0)`
- `17:19:59.959` `launchservicesd`: `QUITTING: pid=66786`

An earlier retained occurrence looks similar:
- `2026-03-16 12:55:03.528` `runningboardd`: Kaku (`pid 23148`) termination reported by launchd `(0, 0, 0)`
- `2026-03-16 12:55:03.522` `WindowServer`: closing Kaku connection for `pid 23148`
- `2026-03-16 12:55:23.660` `launchservicesd`: a new Kaku instance (`pid 34270`) checked in

Why this looks suspicious:
- This does not look like a crash.
- This does not look like an app update/replacement.
[...]

PR #370, opened 2026-05-13
fix(macos): layout-safe Cmd+letter for non-QWERTY keyboards

t0m-car:
Hi! First open-source PR for me, so apologies in advance if I'm missing something.

I'm on a French AZERTY keyboard and noticed Cmd+A would close Kaku entirely, no crash report, just a clean exit. Pretty sure this is what's been hitting #220 too, the quit-origin log you added (`WindowScopeQuitApplication`) is what made the diagnosis possible, so thanks for that.

Cause: on AZERTY the labeled "A" sits on the physical Q position (vkey 12). The Cmd+alnum branch in `key_common` (`window/src/os/macos/window.rs`) forces `prefer_vkey=true` for any alnum vkey, so the event ends up as Cmd+Q downstream, and Cmd+Q is the default QuitApplication binding. Same story for Cmd+Z (CloseCurrentTab) and Cmd+Shift+A (your AI Config shortcut becomes unreachable).

Same shape as the bug you fixed in e485b8a for Cmd+, on AZERTY (punctuation on an alnum vkey), just for letters this time. Added a `is_ascii_letter_text` helper next to your `is_ascii_punctuation_text`, and in `key_common` the Cmd+alnum branch now downgrades `prefer_vkey` to false when `unmod` is a single ASCII letter,  that covers QWERTY / AZERTY / Dvorak / Colemak / Bépo. Empty or non-ASCII `unmod` (real non-Latin IME) keeps the physical-key fallback, so the #147 fix stays intact.

Tested locally on macOS 26.3.1 / Apple Silicon / French AZERTY:

Before: `Cmd+A` → `KeyEvent { key: Char('q'), ... }` → `quit requested origin=WindowScopeQuitApplication`
After: `Cmd+A` → `KeyEvent { key: Char('a'), ... }` → falls through to the shell-side select-all sequence as expected

I deliberately did not add a Cmd+A binding at the app level, saw your call on #261/#262, this PR just stops the existing Cmd+Q default from firing on the wrong key. Cmd+Q on AZERTY (label Q sits on physical A) still quits, as users expect.

Refs #220, #147, #305.

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -3067,6 +3067,11 @@ fn is_ascii_punctuation_text(s: &str) -> bool {
     matches!((chars.next(), chars.next()), (Some(c), None) if c.is_ascii_punctuation())
 }
 
+fn is_ascii_letter_text(s: &str) -> bool {
+    let mut chars = s.chars();
+    matches!((chars.next(), chars.next()), (Some(c), None) if c.is_ascii_alphabetic())
+}
+
 fn is_command_alnum_shortcut(unmod: &str, modifiers: Modifiers, virtual_key: u16) -> bool {
     // Require Cmd (SUPER), disallow Alt/Ctrl (Shift is permitted).
     // Prevents Cmd+alnum shortcuts from failing when a non-Latin IME is active,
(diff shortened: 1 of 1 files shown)


tw93:
@t0m-car, cherry-picked to main (e05dcd4). Great first PR -- thorough analysis and clean fix. AZERTY Cmd+A no longer fires Cmd+Q.

Issue #163, opened 2026-03-09
[FEATURE] Allow disabling Kaku Smart Tab / Tab key override in zsh


qufeng33:
Hi, first of all, thank you for building Kaku. It is a really nice terminal, and I like many of the out-of-the-box integrations and defaults. The overall experience is thoughtful and polished.

I am currently using both Kaku and iTerm2 on the same machine, and both of them share the same zsh configuration. Because of that, I ran into an issue with Kaku's shell integration overriding the `Tab` key behavior globally.

## Problem

Kaku's generated zsh integration installs its own Smart Tab widget and binds `^I` (Tab) to `_kaku_tab_widget`.

In my shell setup, I already use `fzf-tab` for completion UI. After Kaku integration is loaded, `Tab` is no longer owned by `fzf-tab`, so the expected interactive completion UI does not appear anymore.

This is especially noticeable because I use both Kaku and iTerm2. Since they share the same shell config, the Tab override affects my shell behavior outside Kaku as well.

## Expected behavior

It would be great if Kaku could provide an official way to disable Smart Tab / Tab key rebinding in zsh, while keeping the rest of Kaku shell integration enabled.

For example, any of the following would work well:

- an environment variable such as `KAKU_SMART_TAB_DISABLE=1`
- a config option to skip installing `_kaku_tab_widget`
- a config option to skip `bindkey '^I' ...` while keeping other Kaku shell features

## Use case

My current shell stack is roughly:

- zsh
- zsh-autosuggestions
- zsh-completions
- fzf-tab

A concrete example:

1. load my normal zsh config
2. load Kaku shell integration
3. type `cd projects/`
4. press `Tab`

Expected:

- `fzf-tab` shows the interactive completion selector

Actual:

- the Tab behavior is handled by Kaku's Smart Tab path instead

## Why this would help

I think Kaku's Smart Tab is a reasonable default for many users. I am not asking to remove it.

I am only hoping for an official opt-out, so users who already have an established completion workflow can keep using Kaku without changing how `Tab` works across their shared shell environment.

Thanks again for the great work.

tw93:
@qufeng33 Thanks for the clear report. I’ve prepared a fix for this.

Smart Tab now only binds `Tab` inside Kaku sessions by default, and you can still disable it explicitly with `KAKU_SMART_TAB_DISABLE=1`.

I’ve already built a test DMG here:
https://github.com/tw93/static/raw/refs/heads/main/app/Kaku.dmg.zip

[...]

PR #386, opened 2026-05-17
feat: add Smart Tab mode configuration to `kaku config` TUI

### LINKED ISSUE #299: [BUG] 命令查找和补全不完善
在一般情况下，当我们按 Tab键时，系统会自动过滤并补全匹配的命令。但 kaku 工具采用了不同的优先逻辑：它会优先补全最近使用过的命令，只有当没有近期使用记录时，才会回退到补全系统内所有可用的命令。

举个例子：假设我最近使用过 who这条指令，但现在我想用的是 where。如果我只记得这两条命令都以 “wh” 开头，那么当我输入 wh后按下 Tab键，kaku 会直接为我补全为 who，而不会列出所有以 “wh” 开头的选项（如 where）。这就导致我无法通过自动补全找到我真正想用的 where命令。

TopChasing:
## Summary

This PR makes Smart Tab behavior fully configurable from the `kaku config` TUI, and improves documentation to cover all three Smart Tab modes.

## Related Issues

- Related to #163 — Adds the `kaku config` TUI option as a more user-friendly alternative to the `KAKU_SMART_TAB_DISABLE` environment variable
- Related to #299 — The completion-first default introduced to fix #299 is now configurable; users who preferred the old suggestion-first behavior can switch back via `kaku config`

## Motivation

Kaku's Smart Tab supports three modes, but currently:
- Only `KAKU_SMART_TAB_DISABLE` is documented; `KAKU_TAB_ACCEPT_SUGGEST_FIRST` is undocumented
- Both options require manually editing shell environment variables, which is inconvenient for many users
- There is no way to configure Smart Tab from the `kaku config` TUI

This change addresses all three issues.

### 1. `kaku config` TUI — new Smart Tab option (`kaku/src/config_tui/mod.rs`)

Adds a **Smart Tab** field to the Behavior section with three options:

| Option | Behavior |
| :--- | :--- |
| Completion First (default) | Tab shows the completion list; use `→` to accept autosuggestions |
| Suggestion First | Tab accepts autosuggestions when available, falls back to completion |
| Off | Disables Smart Tab entirely, restoring native zsh Tab behavior |

The value is persisted as `config.smart_tab_mode` in `kaku.lua`.

### 2. Lua → shell env var bridge (`assets/macos/Kaku.app/Contents/Resources/kaku.lua`)

Bridges the `config.smart_tab_mode` setting to the shell environment variables (`KAKU_SMART_TAB_DISABLE` / `KAKU_TAB_ACCEPT_SUGGEST_FIRST`) using the existing `config.set_environment_variables` mechanism — the same pattern already used for `COLORFGBG`.

Environment variables set manually in the user's shell rc still take precedence, preserving backward compatibility.

### 3. Documentation (`docs/features.md`)

Expands the existing "Disabling Smart Tab" section into a complete **Smart Tab** configuration guide covering:
- All three modes with a comparison table
- `kaku.lua` configuration examples
- Environment variable fallback for users sharing zshrc across terminals
- Precedence rules

### 4. i18n (`locales/zh-CN.yml`)

Adds Chinese translations for the new config field and option values:
- `smart_tab` → "Smart Tab 模式"
- `completion_first` → "补全优先"
- `suggestion_first` → "建议优先"

## Files Changed

- `kaku/src/config_tui/mod.rs` — Add ConfigField + normalize/serialize logic
- `assets/macos/Kaku.app/Contents/Resources/kaku.lua` — Env var bridge
- `docs/features.md` — Expanded Smart Tab documentation
- `locales/zh-CN.yml` — Chinese translations

## Testing

- The new config field follows the same patterns as existing On/Off/selector fields in the TUI
- The Lua bridge follows the proven `COLORFGBG` pattern at the same code location
- No changes to `setup_zsh.sh` — existing env var handling is reused as-is

### DIFF
--- a/assets/macos/Kaku.app/Contents/Resources/kaku.lua
+++ b/assets/macos/Kaku.app/Contents/Resources/kaku.lua
@@ -3824,6 +3824,9 @@ config.window_close_confirmation = 'SmartPrompt'
 config.tab_close_confirmation = false
 config.pane_close_confirmation = false
 
+-- Smart Tab modes: 'completion_first' (default), 'suggestion_first', or 'off'.
+config.smart_tab_mode = 'completion_first'
+
 -- ===== Tab Bar =====
 config.enable_tab_bar = true
 config.tab_bar_at_bottom = true
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -263,6 +263,12 @@ pub struct Config {
     #[dynamic(default)]
     pub set_environment_variables: HashMap<String, String>,
 
+    /// Controls how the Tab key behaves in zsh inside Kaku sessions.
+    /// The environment variables `KAKU_SMART_TAB_DISABLE` and
+    /// `KAKU_TAB_ACCEPT_SUGGEST_FIRST` are set automatically based on this.
+    #[dynamic(default)]
+    pub smart_tab_mode: SmartTabMode,
+
     /// Specifies the height of a new window, expressed in character cells.
     #[dynamic(default = "default_initial_rows", validate = "validate_row_or_col")]
     pub initial_rows: u16,
@@ -2008,6 +2014,19 @@ impl Config {
             }
         }
 
+        // Smart Tab mode: set the shell env var so setup_zsh.sh picks it up.
+        // User-set env vars in the loop above (from set_environment_variables)
+        // take precedence because they run first and the shell rc can override.
+        match self.smart_tab_mode {
+            SmartTabMode::Off => {
+                cmd.env("KAKU_SMART_TAB_DISABLE", "1");
+            }
+            SmartTabMode::SuggestionFirst => {
+                cmd.env("KAKU_TAB_ACCEPT_SUGGEST_FIRST", "1");
+            }
+            SmartTabMode::CompletionFirst => {}
+        }
+
         if wsl_env.is_some() || cfg!(windows) || crate::version::running_under_wsl() {
             let mut wsl_env = wsl_env.unwrap_or_default();
             if !wsl_env.is_empty() {
@@ -2831,6 +2850,36 @@ impl FromDynamic for BoldBrightening {
                 Ok(false) => Ok(Self::No),
                 Err(_) => Err(err),
             },
+        }
+    }
+}
+
+/// Controls how the Tab key behaves in zsh inside Kaku sessions.
+#[derive(Debug, ToDynamic, Clone, Copy, PartialEq, Eq, Default)]
+pub enum SmartTabMode {
+    /// Tab shows the completion list; use arrow keys to accept autosuggestions.
+    #[default]
+    CompletionFirst,
+    /// Tab accepts autosuggestions when available, falls back to completion.
+    SuggestionFirst,
+    /// Disables Smart Tab entirely, restoring native zsh Tab behavior.
+    Off,
+}
+
+impl FromDynamic for SmartTabMode {
+    fn from_dynamic(
+        value: &wezterm_dynamic::Value,
+        options: wezterm_dynamic::FromDynamicOptions,
+    ) -> Result<Self, wezterm_dynamic::Error> {
+        let s = String::from_dynamic(value, options)?;
+        match s.as_str() {
+            "completion_first" | "CompletionFirst" => Ok(Self::CompletionFirst),
+            "suggestion_first" | "SuggestionFirst" => Ok(Self::SuggestionFirst),
+            "off" | "Off" => Ok(Self::Off),
+            other => Err(wezterm_dynamic::Error::Message(format!(
+                "`{other}` is not a valid SmartTabMode, use one of \
+                 `completion_first`, `suggestion_first`, or `off`"
+            ))),
         }
     }
 }
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -2386,6 +2386,143 @@ mod tests {
             result
         );
     }
+
+    #[test]
+    fn smart_tab_mode_parses_snake_case_strings() {
+        use wezterm_dynamic::{FromDynamic, FromDynamicOptions, Value};
+
+        let options = FromDynamicOptions::default();
+
+        // snake_case variants (written by kaku config TUI)
+        let val = Value::String("completion_first".into());
+        assert_eq!(
+            super::SmartTabMode::from_dynamic(&val, options).unwrap(),
+            super::SmartTabMode::CompletionFirst
+        );
+
+        let val = Value::String("suggestion_first".into());
+        assert_eq!(
+            super::SmartTabMode::from_dynamic(&val, options).unwrap(),
+            super::SmartTabMode::SuggestionFirst
+        );
+
+        let val = Value::String("off".into());
+        assert_eq!(
+            super::SmartTabMode::from_dynamic(&val, options).unwrap(),
+            super::SmartTabMode::Off
+        );
+
+        // PascalCase variants (alternative accepted form)
+        let val = Value::String("CompletionFirst".into());
+        assert_eq!(
+            super::SmartTabMode::from_dynamic(&val, options).unwrap(),
+            super::SmartTabMode::CompletionFirst
+        );
+
+        let val = Value::String("Off".into());
+        assert_eq!(
+            super::SmartTabMode::from_dynamic(&val, options).unwrap(),
+            super::SmartTabMode::Off
+        );
+
+        // Invalid value should error
+        let val = Value::String("invalid_mode".into());
+        assert!(super::SmartTabMode::from_dynamic(&val, options).is_err());
+    }
+
+    #[test]
+    fn smart_tab_mode_field_loads_without_unknown_field_error() {
+        use std::collections::BTreeMap;
+        use wezterm_dynamic::{FromDynamic, FromDynamicOptions, UnknownFieldAction, Value};
+
+        let mut map: BTreeMap<Value, Value> = BTreeMap::new();
+        map.insert(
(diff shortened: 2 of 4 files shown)


tw93:
Hi @TopChasing, thanks for the patch. Reviewed end-to-end and there are two issues that block merging in this form, plus one rebase-time cleanup.

**1. The kaku.lua bridge does not actually fire when users set `config.smart_tab_mode` (blocker).**

The order in a typical user kaku.lua is:

```lua
local config = {}
local ok, loaded = pcall(dofile, bundled)   -- bundled runs HERE, sees config.smart_tab_mode == nil
if ok then config = loaded end
-- TUI inserts the next line just before `return config`:
config.smart_tab_mode = 'off'
return config
```

When `dofile(bundled)` runs, the bundled kaku.lua executes its full body, including the new block at `assets/macos/Kaku.app/Contents/Resources/kaku.lua:4026`:

```lua
local smart_tab = config.smart_tab_mode or 'completion_first'
if smart_tab == 'off' then
  config.set_environment_variables['KAKU_SMART_TAB_DISABLE'] = '1'
elseif smart_tab == 'suggestion_first' then
  config.set_environment_variables['KAKU_TAB_ACCEPT_SUGGEST_FIRST'] = '1'
end
```

at this point `config.smart_tab_mode` is still nil because the user (or `kaku config`) writes the override only after `dofile` returns. The bridge sees `'completion_first'`, no env var is added. The user's `smart_tab_mode = 'off'` lands on the table afterwards, but `set_environment_variables` is already finalized.

The TUI `set_lua_config` confirms this: `kaku/src/config_tui/mod.rs:1388` inserts new fields right before `return config`, i.e. after the dofile. The COLORFGBG line two lines above the new block has the same shape and is similarly stale relative to user-supplied `color_scheme`, which is a small-stakes cosmetic issue; for Smart Tab the stake is the feature itself, so it's worth fixing properly.

The cleanest fix is to do the bridge in Rust, not Lua:

- Add `smart_tab_mode: Option<String>` (or a small enum that derives `FromDynamic`/`ToDynamic`) to `config/src/config.rs`.
- In `Config::apply_cmd_defaults` (`config/src/config.rs:1976`), after the existing `for (k, v) in &self.set_environment_variables` loop, branch on `self.smart_tab_mode` and call `cmd.env(...)` for the appropriate KAKU_* var, but skip if the user's env already has it (mirror the doc claim about shell-rc precedence by checking `std::env::var(...)` or by leaving the user-side `unset` to win the way it does today).

That way the value the user wrote in `kaku.lua` (or via `kaku config`) is what reaches the spawned shell, and no ordering dance is needed.

**2. `config.smart_tab_mode` is not a typed field, so it triggers an unknown-field warning.**

`Config::from_lua` runs with default options, which means `UnknownFieldAction::Warn` on unknown fields (`crates/wezterm-dynamic/src/fromdynamic.rs:35`, `crates/luahelper/src/lib.rs:36`). With `KAKU_STRICT_CONFIG=1` (`config_builder`) the same field hits `UnknownFieldAction::Deny` (`config/src/lua.rs:128`) and the user's whole config fails to load. The fix in (1) resolves this naturally, since adding the field to `Config` registers it as known.

**3. `locales/zh-CN.yml` conflicts (rebase-time cleanup).**

The rust-i18n localization was reverted in b4d779a6, so `locales/zh-CN.yml` no longer exists on main. The yml hunk from this PR needs to be dropped on rebase. The TUI doesn't read i18n strings - `kaku/src/config_tui/mod.rs:458` uses `key: \"Smart Tab\"` literally as the display label - so there's no behavior loss.

What's good to keep:
- TUI field shape, parse/serialize round-trip (mode 1457-1469), and the three-mode option list match the existing patterns in the file.
- Docs in `docs/features.md` are an improvement over today's single-mode reference.

If you'd like, I can pick this up and push the Rust-side bridge on top of your branch; just say the word. Otherwise happy to re-review once (1) is moved into Rust and the yml hunk is dropped.

TopChasing:
> Hi @TopChasing, thanks for the patch. Reviewed end-to-end and there are two issues that block merging in this form, plus one rebase-time cleanup.
> 
> **1. The kaku.lua bridge does not actually fire when users set `config.smart_tab_mode` (blocker).**
> 
> The order in a typical user kaku.lua is:
> 
> ```lua
> local config = {}
> local ok, loaded = pcall(dofile, bundled)   -- bundled runs HERE, sees config.smart_tab_mode == nil
> if ok then config = loaded end
> -- TUI inserts the next line just before `return config`:
> config.smart_tab_mode = 'off'
> return config
> ```
> 
> When `dofile(bundled)` runs, the bundled kaku.lua executes its full body, including the new block at `assets/macos/Kaku.app/Contents/Resources/kaku.lua:4026`:
> 
> ```lua
> local smart_tab = config.smart_tab_mode or 'completion_first'
> if smart_tab == 'off' then
>   config.set_environment_variables['KAKU_SMART_TAB_DISABLE'] = '1'
> elseif smart_tab == 'suggestion_first' then
>   config.set_environment_variables['KAKU_TAB_ACCEPT_SUGGEST_FIRST'] = '1'
> end
> ```
> 
> at this point `config.smart_tab_mode` is still nil because the user (or `kaku config`) writes the override only after `dofile` returns. The bridge sees `'completion_first'`, no env var is added. The user's `smart_tab_mode = 'off'` lands on the table afterwards, but `set_environment_variables` is already finalized.
> 
> The TUI `set_lua_config` confirms this: `kaku/src/config_tui/mod.rs:1388` inserts new fields right before `return config`, i.e. after the dofile. The COLORFGBG line two lines above the new block has the same shape and is similarly stale relative to user-supplied `color_scheme`, which is a small-stakes cosmetic issue; for Smart Tab the stake is the feature itself, so it's worth fixing properly.
> 
> The cleanest fix is to do the bridge in Rust, not Lua:
> 
> * Add `smart_tab_mode: Option<String>` (or a small enum that derives `FromDynamic`/`ToDynamic`) to `config/src/config.rs`.
> * In `Config::apply_cmd_defaults` (`config/src/config.rs:1976`), after the existing `for (k, v) in &self.set_environment_variables` loop, branch on `self.smart_tab_mode` and call `cmd.env(...)` for the appropriate KAKU_* var, but skip if the user's env already has it (mirror the doc claim about shell-rc precedence by checking `std::env::var(...)` or by leaving the user-side `unset` to win the way it does today).
> 
> That way the value the user wrote in `kaku.lua` (or via `kaku config`) is what reaches the spawned shell, and no ordering dance is needed.
> 
> **2. `config.smart_tab_mode` is not a typed field, so it triggers an unknown-field warning.**
> 
> `Config::from_lua` runs with default options, which means `UnknownFieldAction::Warn` on unknown fields (`crates/wezterm-dynamic/src/fromdynamic.rs:35`, `crates/luahelper/src/lib.rs:36`). With `KAKU_STRICT_CONFIG=1` (`config_builder`) the same field hits `UnknownFieldAction::Deny` (`config/src/lua.rs:128`) and the user's whole config fails to load. The fix in (1) resolves this naturally, since adding the field to `Config` registers it as known.
> 
> **3. `locales/zh-CN.yml` conflicts (rebase-time cleanup).**
> 
> The rust-i18n localization was reverted in [b4d779a](https://github.com/tw93/Kaku/commit/b4d779a6f18e0eaca44843a740e7e72d0a034ec0), so `locales/zh-CN.yml` no longer exists on main. The yml hunk from this PR needs to be dropped on rebase. The TUI doesn't read i18n strings - `kaku/src/config_tui/mod.rs:458` uses `key: \"Smart Tab\"` literally as the display label - so there's no behavior loss.
> 
> What's good to keep:
> 
> * TUI field shape, parse/serialize round-trip (mode 1457-1469), and the three-mode option list match the existing patterns in the file.
> * Docs in `docs/features.md` are an improvement over today's single-mode reference.
> 
> If you'd like, I can pick this up and push the Rust-side bridge on top of your branch; just say the word. Otherwise happy to re-review once (1) is moved into Rust and the yml hunk is dropped.

@tw93 Thanks for the detailed review! All three points make sense.

  1. **Lua bridge timing** — you're right, I'll move the bridge to Rust in
  `Config::apply_cmd_defaults` with env-var precedence check.
  2. **Unknown field warning** — will be resolved naturally once it becomes a typed field.
  3. **locales/zh-CN.yml** — my bad, will drop this in the rebase.

  I'd like to fix these myself. Will push an updated commit after rebasing on main.Thanks again for taking the time to write such a thorough review — really helpful for understanding the architecture better. 🙏

tw93:
@TopChasing Thanks a lot for the well-tested PR. Merged in 06dfd67 with the same shape (typed SmartTabMode + kaku config TUI integration + KAKU_SMART_TAB_DISABLE / KAKU_TAB_ACCEPT_SUGGEST_FIRST bridge), plus a small guard so user-set env vars in shell rc still take precedence over the lua config, and the matching kaku.lua wiring. Appreciate the contribution!

TopChasing:
> @TopChasing Thanks a lot for the well-tested PR. Merged in [06dfd67](https://github.com/tw93/Kaku/commit/06dfd6700a1603a62bdf1ad9d9372bfd99eecfe4) with the same shape (typed SmartTabMode + kaku config TUI integration + KAKU_SMART_TAB_DISABLE / KAKU_TAB_ACCEPT_SUGGEST_FIRST bridge), plus a small guard so user-set env vars in shell rc still take precedence over the lua config, and the matching kaku.lua wiring. Appreciate the contribution!

@tw93  Thank you so much! Really appreciate you taking the time to review, refine and merge this — it's a pleasure contributing to such a well-maintained project. I'll keep watching the repo and happy to jump in again anytime! 🚀

PR #429, opened 2026-05-26
fix(deps): unify bitflags and thiserror to single workspace version

darion-yaphet:
## Summary

- Upgrade `workspace.dependencies` from `bitflags = "1.3"` → `"2.11"` and `thiserror = "1.0"` → `"2.0"` to eliminate duplicate compiled versions in the binary
- Point `termwiz`, `wezterm-surface`, `wezterm-escape-parser` back to `bitflags.workspace = true` (they had self-declared `"2.0"` independently)
- Fix all `bitflags!` struct definitions across the workspace to explicitly derive `Clone, Copy, Debug, PartialEq, Eq, Hash` — bitflags 2.x no longer auto-implements these traits
- Extend `FromDynamic`/`ToDynamic` proc-macros to support newtype structs with `#[dynamic(try_from/into)]` (required for `Modifiers`, `WindowDecorations`)
- Fix `Modifiers`/`WindowDecorations` serde to use string conversion path (`serde(into="String", try_from="String")`) instead of raw struct derive, which is incompatible with bitflags 2.x internals
- Fix `.bits` field access → `.bits()` method call in `wezterm-ssh` (bitflags 2.x API change)

## Root Cause

Before this PR, `workspace.dependencies` declared `bitflags = "1.3"` and `thiserror = "1.0"`, but three crates independently declared `bitflags = "2.0"` and two crates declared `thiserror = "2.0"` in their own `Cargo.toml`. Cargo compiled both major versions simultaneously, causing binary bloat and type incompatibility between crates.

## Remaining Duplicates (not fixable here)

`bitflags 1.x` and `thiserror 1.x` still appear transitively via pinned third-party crates (`cocoa = "=0.25.0"`, `libssh-rs`, `cairo-rs`). These are external dependencies and require upstream upgrades.

## Test Plan

- [x] `cargo check` passes with zero errors across all workspace members
- [x] `cargo test` passes with zero failures
- [x] `cargo tree -d` confirms project-owned crates no longer duplicate bitflags or thiserror versions


darion-yaphet:
Closing: branch incorrectly includes fork-only commits. Will reopen with a clean branch based on upstream/main.

PR #430, opened 2026-05-26
fix(deps): unify bitflags and thiserror to single workspace version

darion-yaphet:
## Summary

- Upgrade `workspace.dependencies`: `bitflags 1.3 → 2.11`, `thiserror 1.0 → 2.0`
- Point `termwiz`, `wezterm-surface`, `wezterm-escape-parser` back to `bitflags.workspace = true` (they had self-declared `"2.0"` independently, causing both 1.x and 2.x to be compiled simultaneously)
- Add explicit `Clone, Copy, Debug, PartialEq, Eq, Hash` derives on all `bitflags!` structs (bitflags 2.x no longer auto-implements these)
- Fix `.bits` field access → `.bits()` method in `wezterm-ssh` (bitflags 2.x API change)
- Extend `FromDynamic`/`ToDynamic` proc-macros to support newtype structs with `#[dynamic(try_from/into)]`
- Fix `Modifiers`/`WindowDecorations` serde to use string conversion path

## Test Plan

- [x] `cargo check` passes across all workspace members
- [x] `cargo test` passes with zero failures

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -472,9 +472,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.10.0"
+version = "2.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "812e12b5285cc515a9c72a5c1d3b6d46a19dac5acfef5265968c166106e31dd3"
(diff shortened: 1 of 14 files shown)


tw93:
@darion-yaphet Thanks for the cleanup. I'm going to pass on this one for now: moving bitflags/thiserror to 2.x touches the serde-derive paths that parse user Lua config (`WindowDecorations`, `Modifiers`), and I'd rather not churn those without a concrete need. Appreciate the effort, closing for now and happy to revisit if the 2.x bump becomes necessary.

PR #462, opened 2026-06-15
fix(ai): pass Antigravity CSRF token via stdin not argv

darion-yaphet:
The CSRF token was passed to curl as a -H argument, exposing it in `ps` output to any local user during the request window. Move the token to curl --config - (stdin), matching the existing pattern in fetch_kaku_assistant_models. Payload and non-secret headers stay on argv. Token is escaped for curl config quoting rules.

### DIFF
--- a/kaku/src/ai_config/tui/providers/mod.rs
+++ b/kaku/src/ai_config/tui/providers/mod.rs
@@ -513,13 +513,20 @@ pub(super) fn post_antigravity_lsp_json(
         .ok()?;
     let url = format!("https://127.0.0.1:{https_port}{path}");
 
-    // Token is sent through argv for portability across environments.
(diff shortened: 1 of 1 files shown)


tw93:
Thanks for catching this and sending the patch.

Merged via #462. I kept the Antigravity CSRF-token fix and expanded the same protection to the other usage fetchers: provider Authorization headers and the Claude/Kimi refresh-token form fields now go through `curl --config -` via stdin instead of process argv. I also added tests for curl config escaping and for keeping the Antigravity CSRF token out of argv.

Main checks are green, and I refreshed the notarized nightly build here:
https://github.com/tw93/Kaku/releases/download/nightly/Kaku-nightly.dmg

Please try that build and let me know if anything still looks off.

PR #467, opened 2026-06-17
fix(ai): bypass system proxy for private/loopback hosts

### LINKED ISSUE #466: [BUG] AI requests fail with "error sending request" when a system SOCKS proxy is set
a self-hosted OpenAI-compatible `base_url` on an internal address becomes unreachable whenever a macOS system SOCKS proxy is enabled — even though the server is directly reachable.

## Summary

With a self-hosted OpenAI-compatible `base_url` on an internal address, every AI request fails when a macOS system SOCKS proxy is enabled, even though the server is directly reachable.

## Steps to reproduce

1. Set `base_url` in `assistant.toml` to an internal server, e.g. `http://100.64.0.10:8082/v1` (LAN/RFC1918 or CGNAT/Tailscale `100.64.0.0/10`)
2. Enable a macOS system **SOCKS** proxy (System Settings -> Network -> Proxies -> SOCKS)
3. Run `k "..."` (or open the Cmd+L overlay)

## Expected vs actual

- **Expected**: internal / loopback / LAN hosts connect directly; `NO_PROXY` and the macOS proxy exception list are honored
- **Actual**: `API request failed after 3 attempts: error sending request for url (http://100.64.0.10:8082/v1/chat/completions)`. Yet `curl .../v1/models` returns `200`, and curl even succeeds *through* the same SOCKS proxy.

## Environment

- **Kaku version**: 0.12.2
- **macOS version**: macOS 26.5.1 (25F80)
- **Machine**: MacBook Pro, Apple M5 Max (arm64)
- **Shell**: zsh

## Extra context

Root cause: `build_client_with_proxy()` (`kaku-gui/src/ai_client.rs`) applies the system proxy to every request via `reqwest::Proxy::all()` with no bypass list, and reqwest is built without the `socks` feature, so a SOCKS proxy fails the request at the transport layer. `detect_system_proxy()` also ignores the macOS `scutil` ExceptionsList and the `NO_PROXY` env var, so even internal hosts get force-proxied. A fix is proposed in an accompanying PR.

bluewings1211:
Fixes #466

### Problem

When a macOS system proxy is configured, a self-hosted model server reachable only over loopback / LAN / CGNAT becomes unreachable. `build_client_with_proxy` applied the detected proxy to every AI request via `reqwest::Proxy::all()` with no bypass list, and reqwest is built without the `socks` feature, so a system SOCKS proxy failed the request outright (`error sending request for url ...`) even though the server was directly reachable. `detect_system_proxy()` also ignored the `scutil` ExceptionsList and the `NO_PROXY` env var, so even internal hosts were force-proxied.

### Fix

Attach a reqwest `NoProxy` bypass to the proxied client, covering:

- loopback / link-local / RFC1918 private ranges + `100.64.0.0/10` (CGNAT/Tailscale) + `.local`
- the `NO_PROXY` / `no_proxy` env var
- the macOS `scutil` ExceptionsList

Abbreviated macOS CIDR forms (`169.254/16`) and `*.local` are normalized into reqwest `NoProxy` syntax.

**Files**
- `config/src/proxy.rs` — `system_proxy_exceptions()`, `parse_scutil_exceptions()`, `normalize_no_proxy_entry()` + unit tests
- `kaku-gui/src/ai_client.rs` — `build_no_proxy()` and wiring into `build_client_with_proxy`

**Note:** this intentionally does *not* add reqwest's `socks` feature. Routing an internal `base_url` through a SOCKS proxy is the wrong behavior; bypassing it is the fix. SOCKS support for *external* API hosts can be a separate follow-up.

### Test plan

- [x] `make check`
- [x] `make test` (full workspace via nextest)
- [x] `cargo +nightly fmt --check` clean, `cargo clippy` clean
- [x] 3 new unit tests in `config` (CIDR/exception normalization)
- [x] End-to-end: with a macOS system SOCKS proxy enabled, `k "..."` now reaches an internal server that previously failed with `error sending request`
- [ ] Maintainer to confirm no regression for users who rely on the proxy for *external* API hosts (those still route through the proxy as before)

### DIFF
--- a/config/src/proxy.rs
+++ b/config/src/proxy.rs
@@ -92,6 +92,75 @@ fn is_valid_host(s: &str) -> bool {
     !s.is_empty() && !s.chars().any(|c| c.is_whitespace() || c == '/' || c == ':')
 }
 
+/// Detect the macOS system proxy bypass list (`scutil --proxy` ExceptionsList).
+///
+/// These are the hosts the user marked "Bypass proxy settings for these Hosts"
+/// in Network Settings. Entries are normalized into reqwest `NoProxy` syntax
+/// (partial CIDR masks like `169.254/16` are expanded to `169.254.0.0/16`).
+/// Returns an empty vec on non-macOS or when no exceptions are configured.
+pub fn system_proxy_exceptions() -> Vec<String> {
+    let out = match Command::new("/usr/sbin/scutil").arg("--proxy").output() {
+        Ok(o) if o.status.success() => o,
+        _ => return Vec::new(),
+    };
+    parse_scutil_exceptions(&String::from_utf8_lossy(&out.stdout))
+}
+
+/// Parse the `ExceptionsList` array out of `scutil --proxy` output and
+/// normalize each entry for reqwest `NoProxy::from_string`.
+pub fn parse_scutil_exceptions(text: &str) -> Vec<String> {
+    let mut entries = Vec::new();
+    let mut in_list = false;
+    for line in text.lines() {
+        let trimmed = line.trim();
+        if !in_list {
+            if trimmed.starts_with("ExceptionsList") && trimmed.contains('{') {
+                in_list = true;
+            }
+            continue;
+        }
+        if trimmed.starts_with('}') {
+            break;
+        }
+        // Entries look like `0 : *.local`.
+        if let Some((_, val)) = trimmed.split_once(" : ") {
+            let val = val.trim();
+            if !val.is_empty() {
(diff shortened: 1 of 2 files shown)


tw93:
@bluewings1211 thank you for the fix. I merged #467 as e5e3261, kept the scope to private/loopback no-proxy bypass, and published the follow-up Nightly package so users can try it:

https://github.com/tw93/Kaku/releases/download/nightly/Kaku-nightly.dmg

PR #474, opened 2026-06-29
fix(ai-tools): guard fs_patch against unbounded file read

darion-yaphet:
Summary

fs_patch (the AI filesystem tool that edits a file by replacing old_text with new_text) loaded the entire target file into memory via std::fs::read_to_string with no size check. Its sibling fs_read already caps its read with .take(cap + 512), but fs_patch had no equivalent guard.

When an AI agent directs a patch at a large file inside the project root — a build artifact, log file, dataset, checked-in binary, or minified bundle — this allocates the file's full contents at once and risks an out-of-memory crash of the terminal. For binary files it is even worse: read_to_string buffers the whole file before the UTF-8 validation fails, so the allocation happens regardless. This is a self-inflicted DoS, not a privilege escalation (it requires the agent to target a large file on its own or via prompt injection).

Fix

Add an 8 MiB pre-read size guard in kaku-gui/src/ai_tools/fs.rs:

- MAX_PATCH_FILE_BYTES = 8 MiB — patching needs the whole file in memory (the replacement scans the full content), so unlike fs_read it cannot truncate; the strategy is to refuse oversized files instead.
- ensure_patchable_size(path, limit) — checks metadata().len() and bails before any read allocates the file. Extracted as a helper so the size logic is unit-testable with a small limit.
- Called right before read_to_string in exec_fs_patch.

The 8 MiB ceiling is generous for hand-written source (a 10k-line file is ~300 KB) and only blocks files that should never be edited via per-character replacement. For genuinely large files, the streaming path is shell_exec with sed/perl.

### DIFF
--- a/kaku-gui/src/ai_tools/fs.rs
+++ b/kaku-gui/src/ai_tools/fs.rs
@@ -6,6 +6,30 @@ use std::path::PathBuf;
 
 use super::paths::{reject_if_sensitive, reject_relative_cwd_escape, resolve};
 
+/// Maximum size of a file `fs_patch` will load into memory. Patching needs the
+/// whole file in memory (the replacement scans the full content), so unlike
+/// `fs_read` it cannot truncate. Hand-written source files stay well under this;
+/// the cap exists to refuse multi-hundred-MB build artifacts, logs, datasets, or
+/// checked-in binaries before `read_to_string` allocates their full contents.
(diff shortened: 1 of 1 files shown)


tw93:
@darion-yaphet Thanks for the careful fix and the clear writeup. Merged.

The 8 MiB pre-read guard, and keeping it symmetric with the cap `fs_read` already had, is exactly the right call here, and the unit tests are appreciated. It'll ship in the next release.

PR #476, opened 2026-06-30
refactor(ai-tools): extract shared fs path-validation preamble

darion-yaphet:
## Summary

The six `exec_fs_*` functions in `ai_tools/fs.rs` each duplicated the same four-line preamble — resolve the path arg, then `reject_if_sensitive` followed by `reject_relative_cwd_escape`. The guard order had already drifted from the `mod.rs` dispatcher, which is exactly the failure mode duplicated security checks invite.

## Change

- Add `paths::resolve_checked_arg(args, cwd) -> PathBuf` as the single source of truth for the fs path-validation preamble.
- Collapse the six `exec_fs_*` preambles to one call each (net -21 lines in fs.rs).
- Behavior is unchanged: both guards still run, in the same order.

## Tests

Three new tests in `ai_tools::paths::tests` cover the helper: normal-path resolve, sensitive-path rejection, and missing-`path`-arg error. All existing ai_tools tests pass (paths 22, fs incl. fs_patch), `cargo clippy -p kaku-gui --lib` clean, `cargo fmt` clean.

### DIFF
--- a/kaku-gui/src/ai_tools/fs.rs
+++ b/kaku-gui/src/ai_tools/fs.rs
@@ -4,7 +4,7 @@ use anyhow::{Context, Result};
 use std::io::{BufRead, Read};
 use std::path::PathBuf;
 
-use super::paths::{reject_if_sensitive, reject_relative_cwd_escape, resolve};
+use super::paths::resolve_checked_arg;
 
 /// Maximum size of a file `fs_patch` will load into memory. Patching needs the
 /// whole file in memory (the replacement scans the full content), so unlike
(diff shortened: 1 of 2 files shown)


tw93:
Thanks @darion-yaphet, clean consolidation and the added tests are a nice touch. Landed in 8b5a8af.

The four hand-rolled guard sites in mod.rs (scan/tree/search) still run the two checks in the opposite order. Both must pass either way so the boundary is unchanged, but a raw-path variant of resolve_checked_arg would finish the sweep if you feel like a follow-up.

darion-yaphet:
Thanks, agreed. Since this PR has landed, I will leave the behavior as-is here and take the remaining scan/tree/search guard-order cleanup as a follow-up. A raw-path variant of `resolve_checked_arg` sounds like the right shape to finish that sweep without changing the validation boundary.

PR #485, opened 2026-07-04
feat(tabbar): show foreground process in tab titles

darion-yaphet:
## Summary

  Implements #475.

  - Show stateful foreground processes in tab titles, such as `claude`, `vim`, `lazygit`, and `npm run dev`.
  - Keep idle shell tabs focused on the current directory instead of showing `zsh`, `bash`, or `fish`.
  - Preserve SSH tab behavior by continuing to show the remote target.
  - Apply the same behavior to both the bundled Lua tab formatter and the Rust fallback path.

### DIFF
--- a/assets/macos/Kaku.app/Contents/Resources/kaku.lua
+++ b/assets/macos/Kaku.app/Contents/Resources/kaku.lua
@@ -844,7 +844,7 @@ local function is_shell_foreground(pane)
   end
 
   -- Normalize full executable paths (e.g., /bin/zsh) and login shell names (e.g., -zsh).
-  local name = basename(proc:lower()):gsub("^%-", "")
+  local name = (basename(proc:lower()) or proc:lower()):gsub("^%-", "")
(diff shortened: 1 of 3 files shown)


tw93:
@darion-yaphet Thanks for working on this. I went with a different implementation on main, so I'm closing this PR.

Two issues came up during review:

1. The Lua half never runs in the real app. `format-tab-title` hands the callback PaneInformation objects, which only expose plain fields. On that shape `pane:get_foreground_process_info()` raises (reproduced against mlua 0.9.9, the version Kaku pins), the pcall swallows it, and the helper always returns nil, so tabs keep showing paths.

2. Joining argv tokens for npm/pnpm/yarn/bun can pick up garbage. When a process rewrites its title (node does), KERN_PROCARGS2 leaves environment strings in the argv slots; a real `npm run dev` process reports `argv = ["npm run dev", "COLORFGBG=15;0", ...]`, so the joined title can include env strings.

The version on main reads the `WEZTERM_PROG` user var maintained by shell integration first and falls back to the process executable name. Appreciate the push on #475.

PR #491, opened 2026-07-07
feat(gui): polish tab rename modal

mortalYoung:
## Summary

Polish tab rename modal: pause cursor blink on cursor movement, fix input text alignment.

## Changes

- **Pause cursor blink on movement** — `mark_movement()` records `Instant::now()` on every cursor operation (select all, move left/right, insert/delete chars, paste). Rendering checks elapsed time: cursor stays solid for 500ms after last movement, then resumes normal blink cycle. Uses `Instant` instead of `SystemTime` for monotonic timing.
- **Fix text alignment** — Input area padding changed from `Dimension::Cells` to `Dimension::Pixels(0.0)`, border removed (0px), vertical alignment switched from `Middle` to `Bottom` with adjusted margins for pixel-perfect visual fit.

### DIFF
--- a/kaku-gui/src/termwindow/tab_rename.rs
+++ b/kaku-gui/src/termwindow/tab_rename.rs
@@ -9,7 +9,7 @@ use config::{Dimension, DimensionContext, TabBarColors};
 use mux::tab::TabId;
 use mux::Mux;
 use std::cell::{Ref, RefCell};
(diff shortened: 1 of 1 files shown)


tw93:
@mortalYoung Thanks, merged. I reworked the blink logic on your branch to anchor the whole cycle to the last cursor movement, which keeps the solid-while-typing behavior and avoids the phase jump after the hold expired.

PR #496, opened 2026-07-13
fix(gui): make Cmd+, open single settings window

mortalYoung:
## Summary

- Cmd+, (Open Settings) currently spawns a new settings window every time, piling up duplicates on rapid presses. This fix makes Settings a singleton: if one already exists, Cmd+, focuses it instead of creating another.

- Also fixes a Lua code example in docs where `cwd_obj.host` always get lowercase hostname but `wezterm.hostname` get default hostname.

- Align rename modal text to stay adjacent to the tab item regardless of tab bar position (top/bottom).

## Changes

- **`spawn_command_internal`** returns `anyhow::Result<MuxWindowId>` instead of `()`. Callers that don't need the ID (`spawn_command_impl`, termwindow handler) discard it with the existing `if let Err` pattern — no behavioral change.

- **`ensure_singleton_window(namespace, handler)`** — generic helper in `frontend.rs`. On first call per namespace it runs `handler`, stores the returned `MuxWindowId`. Subsequent calls check if that window is still alive (`Mux::get_window()`) and focus it via `try_front_end → gui_window_for_mux_window → focus`. A `pending` flag guards against duplicate spawns while the first is still in flight.

- **`open_kaku_config`** delegates to `ensure_singleton_window("kaku-config", async || { spawn_command_internal(...).await })`.

- **`docs/configuration.md`** — add `:lower()` to `wezterm.hostname()` comparison in Lua example (`Url::parse` lowercases host per RFC 3986, so the comparison target must also be lowercase to match).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -339,7 +339,7 @@ wezterm.on('augment-command-palette', function(window, pane)
   -- so directories containing spaces or non-ASCII characters work too.
   local host = cwd_obj.host
   if cwd_obj.scheme ~= 'file'
-      or (host and host ~= '' and host ~= 'localhost' and host ~= wezterm.hostname()) then
+      or (host and host ~= '' and host ~= 'localhost' and host ~= wezterm.hostname():lower()) then
     return {}
   end
   local cwd = cwd_obj.file_path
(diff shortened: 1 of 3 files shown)


tw93:
@mortalYoung Thanks, the settings singleton is the right call and the implementation looks good.

I trimmed the branch to the singleton and docs commits. The rename-modal alignment change touches a regression-prone area and needs top/bottom tab bar screenshots to review properly, so please open that one as a separate PR with screenshots attached.

Will merge this once CI is green.

tw93:
Merged, thanks again.

PR #497, opened 2026-07-14
Separate AI chat overlay setup from termwindow key routing

darion-yaphet:
## Summary

  Refactors the `kaku-ai-chat` `EmitEvent` path out of `TermWindow::perform_key_assignment` into a dedicated `termwindow::ai_chat` module.

  The key assignment dispatcher now only routes the event, while the new module owns:

  - AI chat overlay toggling
  - terminal context capture
  - visible line and `@tab` snapshot collection
  - failed command output range handling
  - chat palette mapping
  - overlay layout refresh

  ## Why

  `termwindow/mod.rs` was carrying AI-chat-specific setup inside the general key routing path, which made the dispatcher harder to reason about and increased the local  context needed for future AI chat changes.

  This keeps behavior the same while moving that responsibility behind a smaller boundary.

  ## Verification

  - `cargo fmt --check`
  - `cargo test -p kaku-gui tab_snapshot`
  - `cargo test -p kaku-gui failed_command_output`
  - `make check`

  ## Not Tested

  - `make app` runtime GUI smoke

### DIFF
--- a/kaku-gui/src/termwindow/ai_chat.rs
+++ b/kaku-gui/src/termwindow/ai_chat.rs
@@ -1,0 +1,188 @@
+use super::TermWindow;
+use crate::overlay::start_overlay_pane;
+use mux::pane::{CachePolicy, Pane};
+use std::ops::Range;
+use std::sync::Arc;
+use termwiz::surface::Line;
+use wezterm_term::color::{ColorPalette, SrgbaTuple};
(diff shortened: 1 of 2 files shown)


tw93:
@darion-yaphet Thanks, clean extraction and the extra tests are appreciated. Merged.

PR #498, opened 2026-07-16
Fix shell_exec with fish

shlroland:
## Summary
- build the shell_exec wrapper for fish using fish variable, status, and command-substitution syntax
- preserve the existing POSIX wrapper for bash and zsh
- add regression coverage for both wrapper variants

## Verification
- 
running 2 tests
test ai_tools::shell::tests::shell_exec_wrapper_uses_fish_syntax ... ok
test ai_tools::shell::tests::shell_exec_wrapper_keeps_posix_syntax_for_zsh_and_bash ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 169 filtered out; finished in 0.00s
- verified with fish 4.5.0 that successful and failing commands write cwd and preserve exit codes 0 and 1

### DIFF
--- a/kaku-gui/src/ai_tools/shell.rs
+++ b/kaku-gui/src/ai_tools/shell.rs
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 use std::io::Read;
 use std::os::unix::fs::OpenOptionsExt;
 use std::os::unix::process::CommandExt;
-use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


tw93:
@shlroland Thanks for the fix and the fish verification, this was breaking shell_exec for every fish user. I moved the test module to the end of the file to satisfy clippy, no other changes. Merged.

PR #515, opened 2026-07-31
fix(gui): keep selection drag scroll pinned at scrollback top

F1Justin:
## 简要中文说明

修复鼠标拖拽选择时，光标置于 pane 上/下方触发视口自动滚动，却无限回到原位（视图反复弹回底部）的问题。
根因是 #448 修复引入的 `normalize_viewport` 把低于 scrollback 顶端的滚动位置一律映射为“回到底部”，与拖拽自动滚动冲突；上游 WezTerm 是钳制到顶端。改动：拖拽自动滚动、滚动条轨道 PageUp、copy 模式光标跟随均钳制到 `scrollback_top`；拖拽进行中（`selection_drag_active`）被挤出的视口钉在历史最旧行而不是弹回底部（新增 `drag_normalize_viewport`），拖拽结束仍保留 #448 的跟随输出行为。
新增 2 个单元测试，kaku-gui 全量 425 通过，`cargo +nightly fmt --check` 干净。

## Why

When dragging a text selection with the left button held and the cursor
held above/below the pane, Kaku auto-scrolls the viewport by one row per
mouse event. The scroll works — but whenever the drag reaches the top of
scrollback, the view instantly snaps back to the bottom and the next drag
event starts climbing from the bottom again. The content never stays
scrolled ("无限回到原位" / infinite return to the original position).

Repro (no streaming output needed):

1. Scroll up into scrollback, start a selection drag, keep the cursor
   above the pane edge and keep moving the mouse.
2. Each time the viewport reaches the oldest scrollback row, it jumps
   all the way back to the bottom; the climb restarts from there.

The same snap-back fires mid-drag while output is streaming (e.g.
`tail -f`, Claude Code generating): once the dragged-to content is
pruned from scrollback, the view teleports to the bottom.

## Root cause

`normalize_viewport` maps any position below `scrollback_top` to `None`
(snap to bottom, follow live output). That mapping was introduced by
the #448 fix: when content the user is passively reading is pruned from
scrollback, following live output is less surprising than riding the
moving top edge.

But the mapping is applied indiscriminately — including to *active*
interactive scrolls that merely overshoot the top edge by a row or two.
Upstream WezTerm clamps (`pos.max(dims.scrollback_top)`); Kaku's #448
change turned that clamp into a full teleport to the bottom, which
conflicts with the drag auto-scroll in `extend_selection_at_mouse_cursor`
(`set_viewport(Some(y - 1))`).

## What

Clamp the interactive scroll paths to `scrollback_top` and, while a
selection drag is active, pin a pruned viewport to the oldest remaining
row instead of snapping to the bottom:

- `termwindow/selection.rs` — drag auto-scroll above the pane:
  `set_viewport(Some(y.saturating_sub(1).max(dims.scrollback_top)))`
- `termwindow/mouseevent.rs` — scrollbar track page-up above the thumb:
  same `.max(dims.scrollback_top)` clamp as the wheel path
- `termwindow/mod.rs` — new `drag_normalize_viewport` used by
  `sync_pane_viewport_state` while `selection_drag_active`: a pruned
  viewport pins to `scrollback_top` (when any scrollback remains)
  instead of snapping to the bottom mid-drag. The #448
  follow-live-output behavior is unchanged outside active drags.
- `overlay/copy.rs` — copy-mode cursor-follow scrolls clamped to
  `scrollback_top` (same teleport when the cursor is within
  `vertical_gap` of the oldest row)

## Tests

- New unit tests: `drag_normalize_viewport_pins_pruned_viewport_at_scrollback_top`,
  `drag_normalize_viewport_follows_bottom_when_nothing_left_to_pin`
- Existing viewport normalization tests unchanged and passing
- Full `kaku-gui` suite: 425 passed, 0 failed
- `cargo +nightly fmt --check`: clean
- Manually verified in a debug build: dragging above the pane at the
  top of scrollback now pins instead of teleporting; streaming-output
  drags no longer yank the view back to the bottom mid-drag


tw93:
@F1Justin thank you for tracking this down and for the clear root-cause analysis. I kept your active-drag fix, centralized clamping for all explicit scroll requests, and scoped drag pinning to the captured pane so sibling panes retain the #448 behavior.

I also added regression coverage for Shift+PageUp overshoot, cross-pane drags, and Primary Screen Peek exit. This is now merged into main.

PR #520, opened 2026-08-04
[codex] Follow user Codex provider connections

shlroland:
## What changed

- turn Kaku Assistant's existing Codex auth option into a user-level Codex following mode
- resolve `CODEX_HOME`/`~/.codex` model-provider configuration, authentication, headers, query parameters, retries, and Responses endpoints
- support ChatGPT OAuth, Codex API-key login, environment-backed provider keys, and unauthenticated custom providers
- discover models from the active provider's `/models` endpoint while preserving explicit Kaku Simple/Deep model overrides
- keep Kaku's own conversations, tools, approvals, and agent loop
- fail closed for unsupported authentication and non-Responses providers

## Why

Kaku previously treated Codex as a fixed ChatGPT-login transport. That meant users whose Codex CLI used API-key authentication or a custom Responses-compatible provider could not reuse the same connection in Kaku Assistant.

The root cause was that the Codex path read only the OAuth token and sent every request to a hard-coded ChatGPT backend instead of resolving the user's Codex model-provider configuration.

## User impact

Selecting Codex now follows only the persistent user Codex configuration under `CODEX_HOME`, falling back to `~/.codex`. Project configuration, profiles, and CLI overrides are intentionally excluded. Existing direct API-key, Chat Completions, Responses, and Copilot paths remain available.

## Validation

- `cargo test -p kaku-gui` — 234 library tests and 494 main-binary tests passed; 1 existing test ignored
- `cargo test -p kaku` — 187 tests passed
- AI client regression suite — 47 tests passed
- AI configuration TUI suite — 72 tests passed
- `cargo check -p kaku-gui -p kaku`
- `cargo clippy -p kaku-gui -p kaku -- -D warnings`
- `cargo +nightly fmt --all -- --check`
- `git diff --check`
- `make app`


tw93:
@shlroland Merged, thanks. The failing Clippy run was our own toolchain pin, not your code; it went green once the branch picked up the fixed workflow.

I pushed one commit on top of yours to drop `CONTEXT.md`. The naming rules in it are already carried by the module doc comment and `docs/features.md`, so the repository root did not need a separate glossary. Everything else landed as you wrote it.

The fail-closed handling for unsupported auth and non-Responses providers is the part I liked most: a user whose Codex config points somewhere Kaku cannot follow now finds out before a request goes out rather than after.

This ships in the release after V0.18.0.

PR #532, opened 2026-08-13
[codex] Fix lazygit shortcut in nested terminals

### LINKED ISSUE #531: [BUG] Cmd+Shift+G rejects lazygit in nested terminal working directories
## Problem

`Cmd+Shift+G` can show `Lazygit: Not a git repository` even when the currently interactive shell is inside a valid Git repository.

## Reproduction

1. Start Kaku from a non-repository directory.
2. Start a nested terminal multiplexer such as Herdr or tmux.
3. In its focused Fish pane, use `zi` (or `cd`) to enter a Git repository.
4. Press `Cmd+Shift+G`.

## Actual behavior

Kaku reads the outer pane's cwd and runs `git -C <outer-cwd> rev-parse --show-toplevel`. The outer pane cwd can remain stale while the nested focused Fish pane has a different real `$PWD`, so Kaku blocks dispatch before lazygit runs.

## Expected behavior

Kaku should keep owning the shortcut but dispatch `lazygit` to the active terminal input stream. The receiving shell or nested multiplexer then resolves the command in its actual current directory. Kaku should not use its outer pane cwd as a hard Git precondition.

## Scope

This also covers ordinary directory-changing tools such as `zi`; it is not specific to Herdr. If the final directory is not a Git repository, lazygit should show its own native error.

shlroland:
## Summary

Fixes #531.

`Cmd+Shift+G` could show `Lazygit: Not a git repository` even when the current interactive Fish shell had already entered a valid Git repository with `zi` or `cd`. The problem is not specific to Herdr or tmux: Kaku's cached/OSC 7 pane cwd can lag behind the shell's real `$PWD`, so the old Git preflight checked the wrong directory and blocked dispatch.

The shortcut still belongs to Kaku, but it now checks only that lazygit is installed and sends `Ctrl+U + lazygit + Enter` to the current terminal input stream. The receiving shell or multiplexer resolves the command using its actual focused-pane `$PWD`. If that directory is not a Git repository, lazygit reports the native error itself.

## Changes

- Remove the outer-pane cwd and Git preflight from `launch_lazygit`.
- Preserve the missing-lazygit and dispatch-failure feedback.
- Preserve clear-line-before-send behavior.
- Add a regression guard preventing reintroduction of the outer cwd Git rejection.

## Validation

- `cargo test -p config`
- `make fmt-check`
- `make app`
- `luajit -e "assert(loadfile('assets/macos/Kaku.app/Contents/Resources/kaku.lua'))"`

## Manual verification

1. Start Kaku in a non-Git directory.
2. In a normal Fish pane, use `zi` to enter a Git repository.
3. Confirm `pwd` and `git status` show the repository.
4. Press Cmd+Shift+G.
5. Lazygit should open in the repository selected by Fish, without relying on Kaku's cached pane cwd.

The same flow also covers Herdr and tmux nested panes.

### DIFF
--- a/assets/macos/Kaku.app/Contents/Resources/kaku.lua
+++ b/assets/macos/Kaku.app/Contents/Resources/kaku.lua
@@ -2918,18 +2918,6 @@ local function launch_lazygit(window, pane)
     return
   end
 
-  local path = pane_cwd(pane)
-  if path == "" then
-    show_lazygit_toast(window, pane, "kaku-toast-lazygit-no-cwd")
-    return
-  end
-
-  local repo_root = detect_git_repo_root(path)
-  if not repo_root then
-    show_lazygit_toast(window, pane, "kaku-toast-lazygit-not-git")
-    return
-  end
-
   local lazygit_cmd = resolve_lazygit_command()
   if not lazygit_cmd then
     show_lazygit_toast(window, pane, "kaku-toast-lazygit-missing")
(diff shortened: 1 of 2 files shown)


tw93:
@shlroland Merged as `c7d2c604` and included in the latest Nightly. The active shell now owns the nested cwd decision, so `Cmd+Shift+G` no longer rejects lazygit based on the outer pane path. Exact-main Checks and Build Validation are green. Thanks for the focused fix.

Nightly: https://github.com/tw93/Kaku/releases/download/nightly/Kaku-nightly.dmg

PR #534, opened 2026-08-14
[codex] Preserve alternate-screen UI when clearing terminal

### LINKED ISSUE #533: [BUG] Cmd+K clears alternate-screen TUIs until a tab switch
## Problem

Pressing `Cmd+K` while an alternate-screen TUI is active can make its interface disappear. The interface returns only after switching to another Kaku tab and back.

## Reproduction

1. Start Kaku.
2. Open an alternate-screen TUI such as Herdr (tmux and other full-screen TUIs may also reproduce it).
3. Press `Cmd+K`.

## Actual behavior

Kaku sends `Ctrl+L` and then unconditionally clears both scrollback and the viewport. The second operation erases the TUI's currently rendered alternate-screen viewport. The application may not redraw until a tab switch or another render-triggering event occurs.

## Expected behavior

`Cmd+K` should preserve the visible viewport of an active alternate-screen TUI. It may still discard scrollback. In a normal shell, `Cmd+K` should retain its existing behavior of clearing both the viewport and scrollback.

## Scope

This is a Kaku-side clear-scrollback behavior issue, not a Herdr-specific rendering problem.

shlroland:
## Summary

Fixes #533.

Cmd+K first sends Ctrl+L and then clears both scrollback and the viewport. On an alternate-screen TUI such as Herdr, clearing the viewport erases its current render. The TUI may not repaint until a tab switch triggers another redraw.

This change preserves the viewport when the active pane is on the alternate screen: Cmd+K clears only scrollback in that case. Normal shell panes keep the current behavior of clearing both the viewport and scrollback.

## Changes

- Choose `ScrollbackOnly` instead of `ScrollbackAndViewport` for active alternate-screen panes.
- Preserve the existing clear behavior for normal shell panes.
- Add regression tests for both alternate-screen and normal-screen behavior.

## Validation

- `cargo test -p kaku-gui clear_scrollback_`
- `cargo check -p kaku-gui`
- `make app`

## Manual verification

1. Open Herdr or another alternate-screen TUI in Kaku.
2. Press Cmd+K.
3. The TUI should remain visible immediately, without switching tabs.
4. In a normal shell pane, press Cmd+K and confirm viewport plus scrollback are still cleared.

### DIFF
--- a/kaku-gui/src/termwindow/mod.rs
+++ b/kaku-gui/src/termwindow/mod.rs
@@ -33,7 +33,8 @@ use ::window::*;
 use anyhow::{anyhow, ensure, Context};
 use config::keyassignment::{
     Confirmation, KeyAssignment, LauncherActionArgs, PaneDirection, PaneEncoding, Pattern,
-    PromptInputLine, QuickSelectArguments, RotationDirection, SpawnCommand, SplitSize,
+    PromptInputLine, QuickSelectArguments, RotationDirection, ScrollbackEraseMode, SpawnCommand,
+    SplitSize,
 };
 use config::window::WindowLevel;
 use config::{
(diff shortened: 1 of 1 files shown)


tw93:
@shlroland I landed the accepted fix on `main` as `5fa4e0a5`, preserving your authorship and resolving the conflict with the broadcast removal. The focused regression tests, exact-main CI, and the notarized Nightly are all green, so I am closing this draft as delivered by the equivalent main commit.

Nightly: https://github.com/tw93/Kaku/releases/download/nightly/Kaku-nightly.dmg

PR #536, opened 2026-08-15
feat(ai): recognize OrcaRouter as a named provider

XiaoHuo888-hue:
## Summary

This adds a dedicated [OrcaRouter](https://www.orcarouter.ai) provider rather than relying on the generic OpenAI-compatible endpoint, mirroring how this repo already recognizes Copilot and Codex as named providers. Named routers such as `orcarouter/auto` pick an upstream per request. OrcaRouter is an OpenAI-compatible gateway that exposes 150+ models behind one API key, and it also provides gateway-level security controls for AI agents.

I'm an engineer on the OrcaRouter team.

## What this changes

- `kaku-gui/src/ai_client.rs`: registers `https://api.orcarouter.ai/v1` in `detect_provider_with_auth` so the assistant surfaces "OrcaRouter" as the provider name instead of "Custom", plus unit-test coverage for the new and existing entries.

## Why a named provider rather than the OpenAI-compatible escape hatch

Kaku already lets users point the assistant at any OpenAI-compatible `base_url`, but the provider detection table only names Copilot and Codex; everything else shows up as "Custom". Recognizing OrcaRouter by its endpoint keeps the `/status` readout and config dumps accurate for users on the gateway, the same way Copilot is recognized today. No new dependency and no change to the HTTP path: the existing OpenAI-compatible client handles the transport.

## How to use it

1. Get an API key at https://www.orcarouter.ai/console (keys start with `sk-orca-`).
2. In `~/.config/kaku/assistant.toml` set `base_url = "https://api.orcarouter.ai/v1"` and your `api_key`.
3. Pick a model id such as `openai/gpt-5.5`, `anthropic/claude-sonnet-4.6` or `deepseek/deepseek-v4-pro`. The full catalog is at https://www.orcarouter.ai/models.

## Verification

- Provider detection unit tests pass for the OrcaRouter, Copilot, Codex and Custom cases (including trailing-slash normalization).
- Live API: `POST https://api.orcarouter.ai/v1/chat/completions` with `openai/gpt-5.5` returned HTTP 200 and a valid completion.
- The change is a string-match addition to a pure function with no new imports or type changes, and it is formatted to match the surrounding code. A full workspace `cargo check` was attempted, but the Kaku GUI workspace (a WezTerm-derived tree) is very large and the pinned Rust 1.95.0 toolchain is not usable in the CI-free local sandbox, so the workspace-wide check did not finish within the local time budget.

### DIFF
--- a/kaku-gui/src/ai_client.rs
+++ b/kaku-gui/src/ai_client.rs
@@ -2587,6 +2587,7 @@ fn detect_provider_with_auth(base_url: &str, auth_type: &str) -> &'static str {
     match (normalized.as_str(), auth_type) {
         ("https://api.githubcopilot.com", _) => "Copilot",
         ("https://api.openai.com/v1", "codex") => "Codex",
+        ("https://api.orcarouter.ai/v1", _) => "OrcaRouter",
         _ => "Custom",
     }
 }
@@ -2687,6 +2688,12 @@ mod tests {
             detect_provider_with_auth("https://api.openai.com/v1", "api_key"),
             "Custom"
         );
+        // The OrcaRouter gateway is a named OpenAI-compatible endpoint, so it is
+        // surfaced as its own provider regardless of auth type.
+        assert_eq!(
+            detect_provider_with_auth("https://api.orcarouter.ai/v1", "api_key"),
+            "OrcaRouter"
+        );
         // Unknown / removed providers (Gemini was dropped in V0.10.0) fall
         // through to Custom rather than crashing detection.
         assert_eq!(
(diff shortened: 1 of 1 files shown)


tw93:
@XiaoHuo888-hue Thanks for the clear contribution and disclosure. OrcaRouter already works through Kaku's custom OpenAI-compatible endpoint; I keep named providers for integrations with distinct authentication or transport behavior.

Since this only changes the displayed label without adding functionality, I'm going to keep the generic path and close this as not planned. Thanks again for taking the time.

PR #537, opened 2026-08-17
fix(session): restore pane cwd when OSC 7 reports the local hostname

dufu1991:
## Problem

Session restore reopens tabs/splits correctly, but every pane lands in `$HOME`. `last_session.json` saves each pane's `working_dir`, but `cwd_from_working_dir()` converts it with `url::Url::to_file_path()`, which on unix refuses any host other than empty or `localhost` (verified against the `url` crate source: `None | Some(Host::Domain("localhost")) => None`, `_ => return Err(())`).

OSC 7 is conventionally reported as `file://$HOST$PWD` — WezTerm's own shell integration does exactly this — and the snapshot stores the URL verbatim. Verified live on macOS: emitting `\033]7;file://<hostname>.local/tmp/x\033\\` into a pane makes `kaku cli list` report exactly that URL, and every entry in a real `last_session.json` shows the same shape (`file://df-macbook-pro-14.local/Users/...`). All of them fail `to_file_path()` -> `cwd_for_restore()` returns `None` -> panes spawn at the default cwd.

The ssh-domain branch of `cwd_for_restore()` already special-cases this `to_file_path()` limitation for remote hosts; local panes hit the same wall when the OSC 7 host is the *local* hostname.

## Fix

In `cwd_from_working_dir()`, treat a `file://` URL whose host identifies this machine as local: strip the host before calling `to_file_path()`.

- Accept both the full hostname and its first label (zsh `$HOST`, `hostname`, and `scutil --get LocalHostName` disagree on whether the domain part is included).
- Compare case-insensitively: the URL parser lowercases the host, while macOS hostnames are often mixed case.
- Tolerate a trailing dot (FQDN form).
- Any other host still fails conversion, preserving the existing guarantee that a leftover remote cwd cannot leak into a local spawn.

## Tests

- Extended `restore_cwd_local_domain_requires_local_url` with a `localhost` case; the remote-host rejection case is unchanged.
- Added `restore_cwd_local_domain_accepts_local_hostname`: real hostname (uppercased to prove case-insensitivity), short hostname, and trailing-dot FQDN. The hostname is read at runtime, so the test is machine-independent.

## Verification

- `make fmt-check` passes.
- `cargo test --locked -p kaku-gui session_restore` covers the new and existing restore-cwd tests (cold full-workspace build is slow on the author's machine; the checks workflow covers fmt/check/tests on the PR).

## 中文摘要

问题：会话恢复后所有 pane 都打开在 `$HOME`。快照里保存的 `working_dir` 是带本机真实主机名的 `file://` URL（OSC 7 的惯例格式 `file://$HOST$PWD`），而恢复端 `cwd_from_working_dir()` 调用的 `url::Url::to_file_path()` 在 unix 上拒绝任何非空、非 `localhost` 的 host，导致路径转换失败、pane 以默认目录启动。ssh 分支之前已经针对这个限制做过处理，但本地 pane 在 host 是本机主机名时撞上了同一堵墙。

修复：转换前把"本机主机名"（全名或首段、忽略大小写、容忍 FQDN 尾点）识别为本地并剥离 host；其他 host 依旧拒绝，保留"远程路径不得泄漏进本地 spawn"的防护。附带覆盖本机主机名三种形态的单元测试，并通过 `make fmt-check`。

### DIFF
--- a/kaku-gui/src/local_hostname.rs
+++ b/kaku-gui/src/local_hostname.rs
@@ -1,0 +1,52 @@
+/// Return the current system hostname without assuming it is stable for the
+/// lifetime of the app.
+pub(crate) fn current() -> Option<String> {
+    hostname::get()
+        .ok()
+        .map(|hostname| hostname.to_string_lossy().into_owned())
+        .filter(|hostname| !hostname.is_empty())
+}
+
+fn without_fqdn_dot(hostname: &str) -> &str {
+    hostname.strip_suffix('.').unwrap_or(hostname)
+}
+
+/// Whether a `file://` URL host is proven to identify this machine.
+///
+/// Short hostnames are intentionally not derived from an FQDN: a local
+/// `mac.local` and a remote `mac` are distinct machines even though their
+/// first labels match.
+pub(crate) fn is_local_file_host(host: Option<&str>, local_hostname: Option<&str>) -> bool {
+    let Some(host) = host else {
+        return true;
+    };
+    let host = without_fqdn_dot(host);
+    if host.eq_ignore_ascii_case("localhost") {
+        return true;
+    }
+
+    let Some(local_hostname) = local_hostname else {
+        return false;
+    };
+    let local_hostname = without_fqdn_dot(local_hostname);
+    !local_hostname.is_empty() && host.eq_ignore_ascii_case(local_hostname)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::is_local_file_host;
+
+    #[test]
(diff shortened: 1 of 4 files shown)


tw93:
@dufu1991 Thanks for tracking down the OSC 7 restore case and adding the initial coverage. I tightened hostname validation so a same-named host inside a regular SSH session is not treated as local, and the fix is now merged.

It is available in the latest Nightly: https://github.com/tw93/Kaku/releases/download/nightly/Kaku-nightly.dmg

PR #541, opened 2026-08-28
fix(ai): prefer configured model after provider switch

trxuan:
## Summary

- keep the explicitly configured chat model at the front of a fetched model list
- restore a saved model only when the current provider also advertises it
- replace a stale cross-provider selection with the configured model instead of showing `(list failed)` and selecting the provider's first model

## Root cause

The chat overlay stores `last_model` globally. After switching providers, a saved model from the previous provider may not exist in the new provider's successful `/models` response. The overlay currently treats that mismatch as a failed model fetch and resets `model_index` to the first server model, even when the configured `chat_model` is present.

## Tests

Added regression coverage for:

- a stale model from another provider falling back to the configured model
- a valid saved model remaining selected
- a manually configured model remaining available when omitted from `/models`

### DIFF
--- a/kaku-gui/src/ai_state.rs
+++ b/kaku-gui/src/ai_state.rs
@@ -11,7 +11,7 @@
 //! | Surface | Path | TTL | Purpose |
 //! |---|---|---|---|
 //! | `kaku ai` TUI | `~/.cache/kaku/assistant_models.json` | 30 min, base_url-keyed | Driving the dropdown picker; refreshed every TUI session so a freshly-added model shows up. |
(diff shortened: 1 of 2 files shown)


youxi798:
@trxuan Thanks for the provider-switch fix. I tightened the model state so cached lists stay scoped to the current endpoint, Follow Codex resolves to a discovered model, and background refreshes no longer overwrite an explicit choice.

This is merged and available in the latest Nightly: https://github.com/tw93/Kaku/releases/tag/nightly

PR #552, opened 2026-09-11
fix(ai): make retry backoff responsive to cancellation

yansigit:
Stopping an AI request now interrupts retry backoff promptly. Both HTTP retry paths check cancellation before the first request and after the final sleep, so cancellation does not start another attempt.

SSH behavior stays unchanged: local inline-AI capabilities are not forwarded to remote hosts. The existing agent round limit remains in place, allowing valid repeated reads and polling.

Validation: four retry regression tests passed, including cancelled first requests on both transports; required Clippy checks and formatting passed. The cancellation helper was also checked against the previous implementation to confirm the tests fail before the fix.

### DIFF
--- a/assets/shell-integration/setup_fish.sh
+++ b/assets/shell-integration/setup_fish.sh
@@ -774,44 +774,11 @@ end
 if not functions -q ssh
 function ssh
     set -l _kaku_extra
-    set -l _kaku_ai_capability "$KAKU_AI_INLINE_CAPABILITY"
-    if test -z "$_kaku_ai_capability"; and test -r "$HOME/.config/kaku/ai_inline_capability"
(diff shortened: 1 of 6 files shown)


tw93:
@yansigit Thanks for the retry cancellation improvement. I've merged it with checks before the first request and after the final backoff sleep; both HTTP paths are covered by regression tests.

I left SSH handling unchanged because forwarding the local control capability would let remote hosts authenticate local control messages. I also kept the existing agent round limit, since repeated reads and polling can be valid work. The cancellation fix is available in the [latest Nightly](https://github.com/tw93/Kaku/releases/download/nightly/Kaku-nightly.dmg).

PR #553, opened 2026-09-14
feat(shell): copy the last command transcript with /copy

RichForget:
## Summary
Adds `/copy` at the normal zsh prompt to copy the last completed command and retained output, including start/end timestamps, user/host/directory, and exit status.

- Uses existing semantic command markers and scrollback; no tee, continuous output logging, or renderer changes.
- Handles `/copy` in the zsh line editor, leaving slash commands in foreground terminal applications alone.
- Authenticates the shell-to-GUI clipboard request using the existing capability mechanism.
- Includes a unique request ID so repeated copies work; preserves the clipboard when command boundaries cannot be verified.

## Verification
- Generated zsh integration smoke test: timestamps, nonzero status, repeated requests, and normal `/copy extra` handling.
- `cargo test --locked -p kaku-gui command_copy` — 2 passed.
- `cargo test --locked -p kaku-gui inline_ai_control` — 3 passed.
- `cargo clippy --locked --all-targets -p kaku-gui -- -D warnings` and scoped formatting check passed on the isolated branch.
- Prior local app build and live clipboard checks: successful command, 120 output lines across scrollback, preserved indentation, repeated copy, zero-output command, and exit status 7. Those live checks used the development build before isolating this unchanged feature.

## Scope and limitations
Zsh only; requires updated shell integration and semantic markers. Copies retained terminal content, not overwritten progress frames. Missing or mismatched command boundaries produce a toast without replacing the clipboard. The prompt header is reconstructed as user/host/directory rather than reproducing prompt styling.


tw93:
@RichForget thank you for the careful PR.

Sorry to close this after the work you put in. I am not going to add `/copy` as a prompt command. That would reserve a new word on the shell line and add another privileged clipboard path, and Kaku already has ways to copy the last output.

Select the last command block to copy it; copy-on-select is on by default. Without the mouse, open the command palette with `Cmd+Shift+P` and choose Activate Copy Mode, or press `Ctrl+Shift+X`. If you already know you will need the output, run `that-command | pbcopy`. I am closing this as not planned.

PR #556, opened 2026-09-20
fix(tabbar): draw the tab index on split tabs

TeamMeng:
## What

A tab with a split (two or more panes) renders its pane list without the tab index, even when `show_tab_index_in_tab_bar = true`, so the tab bar looks like the numbering skips around split tabs. Single-pane tabs and the Rust fallback (`build_default_title`) both draw the index.

## Why

`format-tab-title` in the bundled `kaku.lua` returns from the multi-pane branch before the index prefix is applied:

- `assets/macos/Kaku.app/Contents/Resources/kaku.lua:3221` (`if #own_panes > 1 and tab.tab_title == '' then`) returns its `items` early
- the single-pane path only applies `tab_index .. ':'` at `assets/macos/Kaku.app/Contents/Resources/kaku.lua:3395`

Rendered title for a three-pane tab in `apps/psygo` with tab index 2, measured through the same Lua harness the config tests use:

| tab budget | before | after |
| --- | --- | --- |
| 12 | `" apps/psygo "` | `" 2:psygo "` |
| 16 | `" apps/psygo "` | `" 2:apps/psygo "` |
| 32 | `" apps/psygo "` | `" 2:apps/psygo "` |

## How

The index is charged against the same title budget the single-pane path truncates to (`total_limit - leading - 1`, minus the index width), and it is dropped rather than clipped when the tab is too narrow to afford it. The trailing status cell still survives every width budget, which is what the existing narrow-width gate checks.

## Tests

`bundled_kaku_lua_tab_titles_fit_narrow_width_budgets` now enables `show_tab_index_in_tab_bar` in its fixture and asserts a multi-pane title keeps the index whenever the budget can afford it. It fails on `main`:

```
multi-pane title should keep the tab index in a 4-column budget: " r… "
```

- `cargo test -p config`
- `luajit -e "assert(loadfile('assets/macos/Kaku.app/Contents/Resources/kaku.lua'))"` for the top-level 200-locals limit


tw93:
@TeamMeng thank you for the PR.

This is on main. Split tabs now keep the tab index when `show_tab_index_in_tab_bar` is on. It will ship in the next release.

